PR #4, opened 2022-08-23
feat: auction liquidation impl

drbh:
## Security Checklist
WORK IN PROGRESS

- [ ] 1. Re-Entrancy
- [ ] 3. Unexpected Ether
- [ ] 4. Delegatecall
- [ ] 5. Default Visibilities
- [ ] 6. Entropy Illusion

### ⚠️ NOTES ⚠️

### DIFF
--- a/contracts/protocol/libraries/logic/AuctionLiquidationLogic.sol
+++ b/contracts/protocol/libraries/logic/AuctionLiquidationLogic.sol
@@ -1,8 +1,25 @@
 // SPDX-License-Identifier: UNLICENSED
 pragma solidity ^0.8.10;
 
-import {GenericLogic} from "./GenericLogic.sol";
+import {IERC20} from "../../../dependencies/openzeppelin/contracts//IERC20.sol";
+import {GPv2SafeERC20} from "../../../dependencies/gnosis/contracts/GPv2SafeERC20.sol";
+import {PercentageMath} from "../../libraries/math/PercentageMath.sol";
+import {WadRayMath} from "../../libraries/math/WadRayMath.sol";
+import {Helpers} from "../../libraries/helpers/Helpers.sol";
 import {DataTypes} from "../../libraries/types/DataTypes.sol";
+import {ReserveLogic} from "./ReserveLogic.sol";
+import {SupplyLogic} from "./SupplyLogic.sol";
+import {ValidationLogic} from "./ValidationLogic.sol";
+import {GenericLogic} from "./GenericLogic.sol";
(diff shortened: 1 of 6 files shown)


contracts/protocol/libraries/logic/AuctionLiquidationLogic.sol
```diff
@@ -0,0 +1,181 @@
...
+        require(
+            vars.healthFactor < THRESHOLD,
+            "Users health factor must be below the threshold"
+        );
+
+        /// TODO: settle debt from liquidation
+
+        // repay debt
+        // or supply -> wrapped eth gw
+        // also pay fee
```
drbh, line 145: @WalidOfNow can you help point me to an example of how we might supply this back to the market?

contracts/protocol/libraries/logic/AuctionLiquidationLogic.sol
```diff
@@ -0,0 +1,181 @@
...
+            "Bid is too small, please increase"
+        );
+
+        require(
+            params.bidAmount > _auction.highestBid,
+            "Bid is not higher than existing highest bid"
+        );
+
+        // return previous amount to last highest bidder
+        payable(_auction.highestBidder).transfer(_auction.highestBid);
```
drbh, line 169: @WalidOfNow can you help me determine the which contract the ETH should be moved in and out from? Maybe this should be on the pool?

contracts/protocol/pool/Pool.sol
```diff
@@ -486,6 +487,63 @@ contract Pool is ReentrancyGuard, VersionedInitializable, PoolStorage, IPool {
...
+            })
+        );
+    }
+
+    function auctionLiquidationBid(
+        address token,
+        uint256 tokenId,
+        uint256 amount
+    ) external virtual {
+        AuctionLiquidationLogic.executeAuctionLiquidationBid(
```
drbh, line 515: @WalidOfNow should we move the ETH into the pool here?

contracts/protocol/libraries/logic/AuctionLiquidationLogic.sol
WalidOfNow: https://github.com/para-space/paraspace-core/blob/777a606a6a68154fe3882a83226d812b0d3b8d48/contracts/protocol/libraries/logic/LiquidationLogic.sol#L387

contracts/protocol/pool/Pool.sol
WalidOfNow: the eth should has already been sent to the pool and can be checked in the `msg.value`

PR #6, opened 2022-08-24
UniswapV3 fee support in price calculation

WalidOfNow:
## Security Checklist

- [ ] 1. Re-Entrancy
- [ ] 3. Unexpected Ether
- [ ] 4. Delegatecall
- [ ] 5. Default Visibilities
- [ ] 6. Entropy Illusion

### ⚠️ NOTES ⚠️

### DIFF
--- a/contracts/misc/UniswapV3OracleWrapper.sol
+++ b/contracts/misc/UniswapV3OracleWrapper.sol
@@ -2,6 +2,7 @@
 pragma solidity 0.8.10;
 
 import {IAtomicPriceAggregator} from "../interfaces/IAtomicPriceAggregator.sol";
+import {IUniswapV3PositionInfoProvider} from "../interfaces/IUniswapV3PositionInfoProvider.sol";
 import {IParaSpaceOracle} from "../interfaces/IParaSpaceOracle.sol";
 import {IUniswapV3Factory} from "../dependencies/uniswap/IUniswapV3Factory.sol";
 import {IUniswapV3PoolState} from "../dependencies/uniswap/IUniswapV3PoolState.sol";
@@ -10,9 +11,11 @@ import {LiquidityAmounts} from "../dependencies/uniswap/LiquidityAmounts.sol";
 import {TickMath} from "../dependencies/uniswap/libraries/TickMath.sol";
 import {SqrtLib} from "../dependencies/math/SqrtLib.sol";
 import {IERC20Detailed} from "../dependencies/openzeppelin/contracts/IERC20Detailed.sol";
-import {DataTypes} from "../protocol/libraries/types/DataTypes.sol";
 
-contract UniswapV3OracleWrapper is IAtomicPriceAggregator {
+contract UniswapV3OracleWrapper is
+    IAtomicPriceAggregator,
+    IUniswapV3PositionInfoProvider
+{
     IUniswapV3Factory immutable UNISWAP_V3_FACTORY;
     INonfungiblePositionManager immutable UNISWAP_V3_POSITION_MANAGER;
     IParaSpaceOracle immutable PARASPACE_ORACLE;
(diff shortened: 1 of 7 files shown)


contracts/misc/UniswapV3OracleWrapper.sol
```diff
@@ -26,87 +28,251 @@ contract UniswapV3OracleWrapper is IAtomicPriceAggregator {
...
+                    feeGrowthGlobal0X128 -
+                    feeParams.feeGrowthOutside0X128Upper;
+                feeGrowthAbove1X128 =
+                    feeGrowthGlobal1X128 -
+                    feeParams.feeGrowthOutside1X128Upper;
+            }
+            uint256 feeGrowthInside0X128;
+            uint256 feeGrowthInside1X128;
+
+            feeGrowthInside0X128 =
```
GopherJ, line 255: shall we use `Tick.getFeeGrowthInside` instead?

contracts/misc/UniswapV3OracleWrapper.sol
```diff
@@ -26,87 +28,251 @@ contract UniswapV3OracleWrapper is IAtomicPriceAggregator {
...
+                feeGrowthGlobal0X128 -
+                feeGrowthBelow0X128 -
+                feeGrowthAbove0X128;
+            feeGrowthInside1X128 =
+                feeGrowthGlobal1X128 -
+                feeGrowthBelow1X128 -
+                feeGrowthAbove1X128;
+
+            unclaimedTokenAFees =
```
GopherJ, line 264: we can use `FullMath.mulDiv` maybe to be more accurate

contracts/misc/UniswapV3OracleWrapper.sol
zhoujia6139: which `getFeeGrowthInside`?

contracts/misc/UniswapV3OracleWrapper.sol
zhoujia6139: yeah. i think so.

contracts/misc/UniswapV3OracleWrapper.sol
```diff
@@ -141,4 +174,221 @@ contract UniswapV3OracleWrapper is IAtomicPriceAggregator {
...
+
+    function _getLiquidityAmount(
+        uint160 sqrtRatioX96,
+        UinswapV3PositionData memory positionData
+    ) internal pure returns (uint256 token0Amount, uint256 token1Amount) {
+        (token0Amount, token1Amount) = LiquidityAmounts.getAmountsForLiquidity(
+            sqrtRatioX96,
+            TickMath.getSqrtRatioAtTick(positionData.tickLower),
+            TickMath.getSqrtRatioAtTick(positionData.tickUpper),
+            positionData.liquidity
```
GopherJ, line 284: this is the virtual liquidity which cannot reflect users' position, I think we should use `SqrtPriceMath.getAmount0Delta` instead

contracts/misc/UniswapV3OracleWrapper.sol
GopherJ: in v3-core/contracts/libraries/Tick.sol there is `getFeeGrowthInside` function, we can also use this `Tick` library

contracts/misc/UniswapV3OracleWrapper.sol
```diff
@@ -141,4 +174,221 @@ contract UniswapV3OracleWrapper is IAtomicPriceAggregator {
...
+        oracleData.tokenBPrice = PARASPACE_ORACLE.getAssetPrice(
+            positionData.tokenB
+        );
+
+        oracleData.tokenADecimal = IERC20Detailed(positionData.tokenA)
+            .decimals();
+        oracleData.tokenBDecimal = IERC20Detailed(positionData.tokenB)
+            .decimals();
+
+        // TODO using bit shifting for the 2^96
```
GopherJ, line 196: we can use `x << FixedPoint96.RESOLUTION`

GopherJ (approved):

contracts/misc/UniswapV3OracleWrapper.sol
zhoujia6139: we don't have enough parameter info for this function.

PR #10, opened 2022-09-06
Dynamic configs

WalidOfNow:
## Overview

This PR adds a new feature which allows certain NFT collections to have dynamic configurations per token, this is very helpful for UniswapV3 tokens where we can achieve higher capital efficiency while maintaining the security of the protocol (refer to our gitbook for the details on UniswapV3 collateral factor calculations).

The PR introduces a new interface called `IDynamicConfigsStrategy` which has function that returns the dynamic config for a given `tokenId`.  our `GenericLogic` and `LiquidationLogic` can utilize this new interface in order to calculate the dynamic LTV and LT for each token.

This PR also includes a custom Dynamic Configs Strategy for UniswapV3.

## Security Checklist

- [ ] 1. Re-Entrancy
- [ ] 3. Unexpected Ether
- [ ] 4. Delegatecall
- [ ] 5. Default Visibilities
- [ ] 6. Entropy Illusion

### ⚠️ NOTES ⚠️


yz89:
want to know how the code be tested?

contracts/protocol/libraries/logic/GenericLogic.sol
```diff
@@ -129,56 +133,83 @@ library GenericLogic {
...
-                    vars.hasZeroLtvCollateral = true;
-                }
-
-                vars.liquidationThreshold =
-                    vars.userBalanceInBaseCurrency *
-                    (vars.liquidationThreshold);
-
-                vars.avgLiquidationThreshold += vars.liquidationThreshold;
-
                 if (vars.assetType == DataTypes.AssetType.ERC721) {
```
zhoujia6139, line 152: May be we should add a AssetType for UniswapV3 NFT from the very beginning, like `ERC721_UNIV3`. then we can have a more straightforward if else logic like below:
```
if (vars.assetType == DataTypes.AssetType.ERC721_UNIV3) {

} else if (vars.assetType == DataTypes.AssetType.ERC721) {

} else {

}
```

contracts/protocol/libraries/logic/GenericLogic.sol
```diff
@@ -129,56 +133,83 @@ library GenericLogic {
...
+                vars.dynamicConfigs
             ) = currentReserve.configuration.getParams();
 
             unchecked {
                 vars.assetUnit = 10**vars.decimals;
             }
 
+            vars.assetPrice = IPriceOracleGetter(params.oracle).getAssetPrice(
+                vars.currentReserveAddress
+            );
```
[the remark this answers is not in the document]
WalidOfNow: didn't we make it return 0?

contracts/protocol/libraries/logic/GenericLogic.sol
WalidOfNow: the problem with this is that our code won't be as flexible and scalable

contracts/protocol/libraries/logic/GenericLogic.sol
```diff
@@ -194,52 +195,32 @@ library GenericLogic {
...
-                            params.user,
-                            vars.xTokenAddress,
-                            vars.assetUnit,
-                            vars.assetPrice
+                            params,
+                            vars
                         );
-                    }
 
-                    if (!vars.dynamicConfigs) {
```
WalidOfNow: removing this will cause a problem since the dynamicLiquidationThreshold is technically the price of each token multiplied by the its liquidation threshold. did you run the tests after this commit?

zhoujia6139: I moved the check for `vars.dynamicConfigs` to the else branch. and yeah unit test is ok.

contracts/ui/UiPoolDataProvider.sol
```diff
@@ -176,8 +176,9 @@ contract UiPoolDataProvider is IUiPoolDataProvider {
                 reserveData.reserveLiquidationThreshold,
                 reserveData.reserveLiquidationBonus,
                 reserveData.decimals,
-                reserveData.reserveFactor
+                reserveData.reserveFactor,
                 // eModeCategoryId
```
GopherJ, line 180: missing `getUniswapV3LpTokenData` change

zhoujia6139 (approved):

PR #11, opened 2022-09-06
fix: ntoken collatoral balance

zhoujia6139:
## Security Checklist

- [ ] 1. Re-Entrancy
- [ ] 3. Unexpected Ether
- [ ] 4. Delegatecall
- [ ] 5. Default Visibilities
- [ ] 6. Entropy Illusion

### ⚠️ NOTES ⚠️


zhoujia6139:
Need refactor after review

zhoujia6139:
### for supply issue, this is a bug:
if Alice  supply 1 punk with useAsCollateral=false
then Alice supply another punk with useAsCollateral = true
then Alice's usingAsCollateral flag for Punk is false, but Alice want to collatoralise one of her punk.

### for withdraw issue, this need to optimize:
if Alice have supplied 1 punk and not use it as collateral.
so Alice's usingAsCollateral flag for Punk is false and collaterizedBalance is 0
then when she withdraw the punk, we will reset Alice's usingAsCollateral flag for Punk to false, but no need to do it.

### for refactor executeUseReserveAsCollateral and executeUseERC721AsCollateral
1, when Alice have supplied 1 punk and use it as collateral. 
so Alice's usingAsCollateral flag for Punk is true and collaterizedBalance is 1, 
when she call executeUseReserveAsCollateral to uncollatoralise her punk, 
we only set Alice's usingAsCollateral flag for Punk to false but leave the collaterizedBalance to 1

2, there is no interface if user want to collatoralise or uncollatoralise multi tokenid but not all.

### for the liquidation issue:
if Alice have supplied 1 punk and Alice use it as collateral to borrow some other asset.
when hf < 1, Alice liquidate herself, 
this will cause Alice's usingAsCollateral flag for Punk is false and collaterizedBalance is 0


### for the finalTransfer issue, this is a serious issue:
if Alice have supplied 1 punk which tokenid is 0, and Alice use it as collateral to borrow some other asset.
then Alice withdraw the punk and leave the debt.

contracts/protocol/libraries/logic/ValidationLogic.sol
```diff
@@ -593,11 +597,8 @@ library ValidationLogic {
     }
 
     function validateSetUseERC721AsCollateral(
-        DataTypes.ReserveCache memory reserveCache,
-        address sender,
-        address owner
+        DataTypes.ReserveCache memory reserveCache
     ) internal pure {
-        require(sender == owner, Errors.NOT_THE_OWNER);
```
GopherJ, line 600: missing assetType check

contracts/protocol/libraries/logic/ValidationLogic.sol
```diff
@@ -700,6 +701,11 @@ library ValidationLogic {
             Errors.INVALID_ASSET_TYPE
         );
 
+        require(
+            params.liquidator != params.borrower,
```
GopherJ, line 705: why do we need to pass `liquidator` instead of just using msg.sender?

contracts/protocol/libraries/logic/SupplyLogic.sol
```diff
@@ -165,12 +165,11 @@ library SupplyLogic {
             );
         }
 
-        bool isFirstSupply = INToken(reserveCache.xTokenAddress).mint(
+        bool isFirstCollateral = INToken(reserveCache.xTokenAddress).mint(
```
GopherJ, line 168: actually I think `mint` can return back:

```
(oldCollateralBalance, newCollateralBalance)
```

which can be used in any case

GopherJ:
```
oldCollateralBalance == 0 && newCollateralBalance > oldCollateralBalance => set true
oldCollateralBalance != 0 && newCollateralBalance == 0 => set false
```

contracts/protocol/libraries/logic/SupplyLogic.sol
```diff
@@ -203,12 +202,11 @@ library SupplyLogic {
             DataTypes.AssetType.ERC721
         );
 
-        bool isFirstSupply = INToken(reserveCache.xTokenAddress).mint(
+        bool isFirstCollateral = INToken(reserveCache.xTokenAddress).mint(
```
GopherJ, line 205: `oldCollateralBalance == 0 && newCollateralBalance > oldCollateralBalance`

contracts/protocol/libraries/logic/SupplyLogic.sol
```diff
@@ -203,12 +202,11 @@ library SupplyLogic {
...
             params.onBehalfOf,
             params.tokenData
         );
-        // TODO consider using (usedAsCollateral > 0) instead here to enable collateralization
-        if (isFirstSupply) {
+        if (isFirstCollateral) {
```
GopherJ, line 209: `oldCollateralBalance == 0 && newCollateralBalance > oldCollateralBalance`

contracts/protocol/libraries/logic/SupplyLogic.sol
```diff
@@ -504,45 +495,54 @@ library SupplyLogic {
...
-            bool valid,
-            address owner,
-            uint256 collaterizedBalance
-        ) = ICollaterizableERC721(reserveCache.xTokenAddress)
-                .setIsUsedAsCollateral(tokenId, useAsCollateral);
+        bool isAnyValid = false;
+        address sender = msg.sender;
+        for (uint256 index = 0; index < tokenIds.length; index++) {
+            bool valid = ICollaterizableERC721(reserveCache.xTokenAddress)
+                .setIsUsedAsCollateral(
```
GopherJ, line 510: we should wrap batch operation inside NToken

contracts/protocol/libraries/logic/SupplyLogic.sol
```diff
@@ -504,45 +495,54 @@ library SupplyLogic {
...
-                owner
-            );
+        if (isAnyValid) {
+            ValidationLogic.validateSetUseERC721AsCollateral(reserveCache);
 
             if (useAsCollateral) {
-                if (collaterizedBalance == 1) {
+                // here we just need to check flag and don't need to check collateral balance
+                // because collateral balance must > 0.
+                if (!userConfig.isUsingAsCollateral(reserve.id)) {
```
GopherJ, line 526: `oldCollateralBalance == 0 && newCollateralBalance > oldCollateralBalance`

contracts/protocol/libraries/logic/SupplyLogic.sol
```diff
@@ -504,45 +495,54 @@ library SupplyLogic {
...
                 }
-                // TODO emit event
             } else {
-                if (collaterizedBalance == 0) {
+                // here we just need to check collateral balance and don't need to check flag
+                // because isUsingAsCollateral flag must be true
+                uint256 userBalance = ICollaterizableERC721(
+                    reserveCache.xTokenAddress
+                ).collaterizedBalanceOf(sender);
+                if (userBalance == 0) {
```
GopherJ, line 536: `oldCollateralBalance != 0 && newCollateralBalance == 0`

contracts/protocol/libraries/logic/ValidationLogic.sol
zhoujia6139: is there any difference?

contracts/protocol/libraries/logic/SupplyLogic.sol
zhoujia6139: but obviously `isFirstCollateral` has a simpler check logic and better understanding for reading code.

contracts/protocol/libraries/logic/SupplyLogic.sol
zhoujia6139: same as above. `isFirstCollateral` is nice and simpler.

contracts/protocol/libraries/logic/SupplyLogic.sol
```diff
@@ -504,45 +495,54 @@ library SupplyLogic {
...
+            }
+        }
 
-        if (valid) {
-            ValidationLogic.validateSetUseERC721AsCollateral(
-                reserveCache,
-                msg.sender,
-                owner
-            );
+        if (isAnyValid) {
```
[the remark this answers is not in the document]
zhoujia6139: if isAnyValid is true means `oldCollateralBalance != newCollateralBalance`.  is there any difference?

contracts/protocol/libraries/logic/SupplyLogic.sol
GopherJ: but it's more general to handle more cases. You can also easily wrap it inside a helper function

contracts/protocol/libraries/logic/SupplyLogic.sol
zhoujia6139: yeah. we can add a function to support batch operation.

contracts/protocol/libraries/logic/SupplyLogic.sol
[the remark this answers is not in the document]
GopherJ: difference is that I still want NToken to provide this basic info,

is any valid 

SupplyLogic is already heavy

GopherJ (approved):

PR #12, opened 2022-09-07
feat: add ETH support for Increase and Decrease Liquidity by NTokenUn…

zhoujia6139:
## Security Checklist

- [ ] 1. Re-Entrancy
- [ ] 3. Unexpected Ether
- [ ] 4. Delegatecall
- [ ] 5. Default Visibilities
- [ ] 6. Entropy Illusion

### ⚠️ NOTES ⚠️

### DIFF
--- a/contracts/protocol/tokenization/NTokenUniswapV3.sol
+++ b/contracts/protocol/tokenization/NTokenUniswapV3.sol
@@ -89,7 +89,7 @@ contract NTokenUniswapV3 is NToken {
         uint128 liquidityDecrease,
         uint256 amount0Min,
         uint256 amount1Min,
-        bool recevieEthAsWeth
+        bool receiveEthAsWeth
     ) internal returns (uint256 amount0, uint256 amount1) {
         if (liquidityDecrease > 0) {
             // amount0Min and amount1Min are price slippage checks
@@ -125,12 +125,12 @@ contract NTokenUniswapV3 is NToken {
         ) = INonfungiblePositionManager(_underlyingAsset).positions(tokenId);
 
         address weth = _addressesProvider.getWETH();
-        recevieEthAsWeth = (recevieEthAsWeth && (token0 == weth || token1 == weth));
+        receiveEthAsWeth = (receiveEthAsWeth && (token0 == weth || token1 == weth));
 
         INonfungiblePositionManager.CollectParams
             memory collectParams = INonfungiblePositionManager.CollectParams({
                 tokenId: tokenId,
-                recipient: recevieEthAsWeth ? address(this) : _msgSender(),
+                recipient: receiveEthAsWeth ? address(this) : _msgSender(),
                 amount0Max: 2**128 - 1,
                 amount1Max: 2**128 - 1
             });
@@ -141,7 +141,7 @@ contract NTokenUniswapV3 is NToken {
         require(amount0>=amount0Min, "Insufficient amount0");
         require(amount1>=amount1Min, "Insufficient amount1");
 
-        if (recevieEthAsWeth) {
+        if (receiveEthAsWeth) {
             uint256 balanceWeth = IERC20(weth).balanceOf(address(this));
             if (balanceWeth > 0) {
                 IWETH(weth).withdraw(balanceWeth);
@@ -306,13 +306,13 @@ contract NTokenUniswapV3 is NToken {
         uint128 liquidityDecrease,
         uint256 amount0Min,
         uint256 amount1Min,
-        bool recevieEthAsWeth
+        bool receiveEthAsWeth
     ) external {
         // only the token owner of the NToken can decrease the underlying
         require(_msgSender() == ownerOf(tokenId), Errors.NOT_THE_OWNER);
 
         // interact with Uniswap V3
-        _decreaseLiquidity(tokenId, liquidityDecrease, amount0Min, amount1Min, recevieEthAsWeth);
+        _decreaseLiquidity(tokenId, liquidityDecrease, amount0Min, amount1Min, receiveEthAsWeth);
 
         // return data about the users healthFactor after decrease
         (
--- a/contracts/protocol/tokenization/NTokenUniswapV3.sol
+++ b/contracts/protocol/tokenization/NTokenUniswapV3.sol
@@ -14,6 +14,7 @@ import {IPool} from "../../interfaces/IPool.sol";
 import {NToken} from "./NToken.sol";
 import {DataTypes} from "../libraries/types/DataTypes.sol";
 import {INonfungiblePositionManager} from "../../dependencies/uniswap/INonfungiblePositionManager.sol";
+import {IWETH} from "../../misc/interfaces/IWETH.sol";
 
 /**
  * @title UniswapV3 NToken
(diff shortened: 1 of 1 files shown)


contracts/protocol/tokenization/NTokenUniswapV3.sol
```diff
@@ -107,44 +109,73 @@ contract NTokenUniswapV3 is NToken {
...
             });
 
         (amount0, amount1) = INonfungiblePositionManager(_underlyingAsset)
             .collect(collectParams);
+
+        require(amount0 >= amount0Min, "Insufficient amount0");
+        require(amount1 >= amount1Min, "Insufficient amount1");
+
+        if (receiveEthAsWeth) {
+            uint256 balanceWeth = IERC20(weth).balanceOf(address(this));
```
GopherJ, line 146: quick return is better than nested scope

contracts/protocol/tokenization/NTokenUniswapV3.sol
```diff
@@ -107,44 +109,73 @@ contract NTokenUniswapV3 is NToken {
...
+            ,
+            ,
+            ,
+            ,
+            ,
+            ,
+
+        ) = INonfungiblePositionManager(_underlyingAsset).positions(tokenId);
+
+        address weth = _addressesProvider.getWETH();
```
GopherJ, line 127: we start to have multiple cases for weth wording, I think we can be consistent and use only:

WETH because it's a keyword

instead of having:

- weth
- Weth
- WETH

contracts/protocol/tokenization/NTokenUniswapV3.sol
```diff
@@ -107,44 +109,73 @@ contract NTokenUniswapV3 is NToken {
...
             .collect(collectParams);
+
+        require(amount0 >= amount0Min, "Insufficient amount0");
+        require(amount1 >= amount1Min, "Insufficient amount1");
+
+        if (receiveEthAsWeth) {
+            uint256 balanceWeth = IERC20(weth).balanceOf(address(this));
+            if (balanceWeth > 0) {
+                IWETH(weth).withdraw(balanceWeth);
+                (bool success, ) = _msgSender().call{value: balanceWeth}(
```
GopherJ, line 149: re-entrancy guard will be needed, is `gas limit` necessary?

GopherJ: I think in our code base we have also `safeTransfer` for ETH, can we reuse?

contracts/protocol/tokenization/NTokenUniswapV3.sol
```diff
@@ -161,31 +192,42 @@ contract NTokenUniswapV3 is NToken {
...
-        if (token1Allownance == 0) {
-            IERC20(token1).safeApprove(_underlyingAsset, 2**256 - 1);
+        //avoid stack too deep
+        {
+            uint256 token0Allownance = IERC20(token0).allowance(
+                address(this),
+                _underlyingAsset
+            );
+            if (token0Allownance == 0) {
+                IERC20(token0).safeApprove(_underlyingAsset, 2**256 - 1);
```
GopherJ, line 221: 2 ** 256 can be pre-computed

contracts/protocol/tokenization/NTokenUniswapV3.sol
```diff
@@ -161,31 +192,42 @@ contract NTokenUniswapV3 is NToken {
...
         }
 
-        uint256 token1Allownance = IERC20(token1).allowance(
-            address(this),
-            _underlyingAsset
-        );
-        if (token1Allownance == 0) {
-            IERC20(token1).safeApprove(_underlyingAsset, 2**256 - 1);
+        //avoid stack too deep
+        {
```
GopherJ, line 215: we can define a LocalVars to avoid this block scope

contracts/protocol/tokenization/NTokenUniswapV3.sol
```diff
@@ -161,31 +192,42 @@ contract NTokenUniswapV3 is NToken {
...
+            if (token0Allownance == 0) {
+                IERC20(token0).safeApprove(_underlyingAsset, 2**256 - 1);
+            }
+
+            uint256 token1Allownance = IERC20(token1).allowance(
+                address(this),
+                _underlyingAsset
+            );
+            if (token1Allownance == 0) {
+                IERC20(token1).safeApprove(_underlyingAsset, 2**256 - 1);
```
GopherJ, line 229: this is duplicated with token0, can we define a common function?

contracts/protocol/tokenization/NTokenUniswapV3.sol
```diff
@@ -201,20 +243,32 @@ contract NTokenUniswapV3 is NToken {
...
         }
 
-        if (amount1 < amountAdd1) {
+        if (amount1 < amountAdd1 && !token1IsETH) {
             uint256 refund1 = amountAdd1 - amount1;
             IERC20(token1).safeTransfer(_msgSender(), refund1);
         }
+
+        //refund eth
+        if (txValue > 0) {
```
GopherJ, line 262: quick return is better to avoid nested scope

contracts/protocol/tokenization/NTokenUniswapV3.sol
```diff
@@ -201,20 +243,32 @@ contract NTokenUniswapV3 is NToken {
...
+            INonfungiblePositionManager(_underlyingAsset).refundETH();
+            uint256 ethBalance = address(this).balance;
+            if (ethBalance > 0) {
+                (bool success, ) = msg.sender.call{value: ethBalance}(
```
GopherJ, line 266: I think we can use safeTransferETH, I remember we have it in our code base

GopherJ (commented):
as long as the refactor is done, we can merge. Currently it's still composed of many duplicated transfer/approve/nested scope etc

GopherJ (approved):

contracts/protocol/tokenization/NTokenUniswapV3.sol
zhoujia6139: ok, it can be replaced with `0xffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff`.

contracts/protocol/tokenization/NTokenUniswapV3.sol
zhoujia6139: I think we don't need re-entrancy guard here because UniswapV3Pool control the position state and there is no erc20 asset on this contract. NonfungiblePositionManager also don't have any re-entrancy guard.

contracts/protocol/tokenization/NTokenUniswapV3.sol
zhoujia6139: yeah.

contracts/protocol/tokenization/NTokenUniswapV3.sol
zhoujia6139: I think it's ok, we can have the style like this.
```
    IWETH internal immutable WETH;
    address public immutable weth;
```

PR #16, opened 2022-09-10
feat: auction based liquidation

GopherJ:

## Security Checklist

- [ ] 1. Re-Entrancy
- [ ] 3. Unexpected Ether
- [ ] 4. Delegatecall
- [ ] 5. Default Visibilities
- [ ] 6. Entropy Illusion

### ⚠️ NOTES ⚠️


contracts/ui/UiPoolDataProvider.sol
```diff
@@ -260,29 +261,60 @@ contract UiPoolDataProvider is IUiPoolDataProvider {
...
+                nTokenAddresses.length
+            );
+        IPool pool = IPool(provider.getPool());
+
+        for (uint256 i = 0; i < nTokenAddresses.length; i++) {
+            address asset = nTokenAddresses[i];
+            uint256 userTotalBalance = INToken(asset).balanceOf(user);
+            tokenData[i] = new DataTypes.AuctionData[](userTotalBalance);
+
+            for (uint256 j = 0; j < userTotalBalance; j++) {
```
jfzhou5, line 285: I think we can use `tokenIds[i].length` as the loop count instead of `userTotalBalance `, If we use `userTotalBalance`, userTotalBalance must be less than  `tokenIds[i].length`.

contracts/protocol/pool/DefaultReserveAuctionStrategy.sol
```diff
@@ -0,0 +1,136 @@
...
+        return _stepExp;
+    }
+
+    function getTickLength() external view returns (uint256) {
+        return _tickLength;
+    }
+
+    function calculateAuctionPriceMultiplier(
+        uint256 auctionStartTimestamp,
+        uint256 currentTimestamp
```
WalidOfNow, line 92: do you need to pass the current timestamp? seems unnecessary

contracts/protocol/pool/PoolConfigurator.sol
```diff
@@ -186,6 +188,29 @@ contract PoolConfigurator is VersionedInitializable, IPoolConfigurator {
         );
     }
 
+    /// @inheritdoc IPoolConfigurator
+    function configureReserveAsAuctionCollateral(
+        address asset,
+        bool auctionEnabled,
+        uint256 auctionRecoveryHealthFactor
```
WalidOfNow, line 195: why are we setting this here? can't we set it in the `setAuctionConfiguration` ?

PR #19, opened 2022-09-14
fix: should check hasAnyCollateralAsset before burn nToken

zhoujia6139:
## Security Checklist

- [ ] 1. Re-Entrancy
- [ ] 3. Unexpected Ether
- [ ] 4. Delegatecall
- [ ] 5. Default Visibilities
- [ ] 6. Entropy Illusion

### ⚠️ NOTES ⚠️

### DIFF
--- a/contracts/protocol/libraries/logic/SupplyLogic.sol
+++ b/contracts/protocol/libraries/logic/SupplyLogic.sol
@@ -310,16 +310,9 @@ library SupplyLogic {
         DataTypes.ReserveData storage reserve = reservesData[params.asset];
         DataTypes.ReserveCache memory reserveCache = reserve.cache();
 
+        ValidationLogic.validateWithdrawERC721(reserveCache);
         reserve.updateState(reserveCache);
         uint256 amountToWithdraw = params.tokenIds.length;
-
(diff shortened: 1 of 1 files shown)


contracts/protocol/libraries/logic/SupplyLogic.sol
```diff
@@ -310,17 +310,10 @@ library SupplyLogic {
         DataTypes.ReserveData storage reserve = reservesData[params.asset];
         DataTypes.ReserveCache memory reserveCache = reserve.cache();
 
+        ValidationLogic.validateWithdrawERC721(reserveCache);
```
GopherJ, line 313: should be after      
```
   reserve.updateState(reserveCache);
```

normally although it's erc721 reserve

GopherJ:
@zhoujia6139 can this be merged into #11  because it's similar thing, overall any change regarding collateral balance will need to be well designed and tested so it'll take time to merge

zhoujia6139:
yeah. these two pr do have some little connection. but this one is more simple and urgent, if there are no more question we can merge it soon.

GopherJ (approved):

PR #35, opened 2022-09-23
adjust Pool.RescueTokens for ERC721

jfzhou5:
## Security Checklist

- [ ] 1. Re-Entrancy
- [ ] 3. Unexpected Ether
- [ ] 4. Delegatecall
- [ ] 5. Default Visibilities
- [ ] 6. Entropy Illusion

### ⚠️ NOTES ⚠️

### DIFF
--- a/contracts/protocol/pool/Pool.sol
+++ b/contracts/protocol/pool/Pool.sol
@@ -992,10 +992,11 @@ contract Pool is ReentrancyGuard, VersionedInitializable, PoolStorage, IPool {
     /// @inheritdoc IPool
     function rescueTokens(
         address token,
+        DataTypes.AssetType assetType,
         address to,
-        uint256 amount
+        uint256 amountOrTokenId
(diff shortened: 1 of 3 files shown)


contracts/protocol/pool/Pool.sol
```diff
@@ -992,10 +992,11 @@ contract Pool is ReentrancyGuard, VersionedInitializable, PoolStorage, IPool {
     /// @inheritdoc IPool
     function rescueTokens(
         address token,
+        DataTypes.AssetType assetType,
```
GopherJ, line 995: could you put assetType before token which makes more sense. and this may breaks some test cases in paraspace-periphery

jfzhou5: have adjusted the order of the params. There are no breaks for test cases, And I have added new unit tests for rescueTokens function.

zhoujia6139 (approved):

PR #41, opened 2022-09-26
Protocol improvements (Pool Proxy, finalizeTransfer, collateralizedBalance, AssetType, DynamicConfig)

WalidOfNow:
## Security Checklist

- [ ] 1. Re-Entrancy
- [ ] 3. Unexpected Ether
- [ ] 4. Delegatecall
- [ ] 5. Default Visibilities
- [ ] 6. Entropy Illusion

### ⚠️ NOTES ⚠️


contracts/protocol/tokenization/base/MintableIncentivizedERC721.sol
```diff
@@ -642,15 +642,16 @@ abstract contract MintableIncentivizedERC721 is
     function _transferCollaterizable(
         address from,
         address to,
-        uint256 tokenId,
-        bool isUsedAsCollateral
-    ) internal virtual {
-        MintableIncentivizedERC721._transfer(from, to, tokenId);
+        uint256 tokenId
+    ) internal virtual returns (bool isUsedAsCollateral) {
+        isUsedAsCollateral = _isUsedAsCollateral[tokenId];
```
GopherJ, line 647: what if from == to?

contracts/protocol/pool/PoolCore.sol
```diff
@@ -656,6 +624,37 @@ contract PoolCore is
...
+        SupplyLogic.executeFinalizeTransferERC721(
+            _reserves,
+            _reservesList,
+            _usersConfig,
+            DataTypes.FinalizeTransferParams({
+                asset: asset,
+                from: from,
+                to: to,
+                usedAsCollateral: usedAsCollateral,
+                value: 1,
```
GopherJ, line 649: value is too general I think we can rename to amount

contracts/protocol/tokenization/base/MintableIncentivizedERC721.sol
GopherJ: also need to change _liquidateNTokens if we really want to do so

contracts/protocol/libraries/logic/SupplyLogic.sol
```diff
@@ -446,6 +435,56 @@ library SupplyLogic {
...
+                        reservesList,
+                        usersConfig[params.from],
+                        params.asset,
+                        params.from,
+                        params.reservesCount,
+                        params.oracle
+                    );
+                }
+
+                if (params.balanceFromBefore == 1) {
```
GopherJ, line 477: executeCollateralizedERC721 also needs to be updated

GopherJ: ok maybe not necessary, finalizeTransfer only useful for transfer

contracts/protocol/tokenization/base/MintableIncentivizedERC721.sol
```diff
@@ -642,15 +642,16 @@ abstract contract MintableIncentivizedERC721 is
...
 
-        if (isUsedAsCollateral) {
+        if (from != to && isUsedAsCollateral) {
             _userState[from].collaterizedBalance -= 1;
```
zhoujia6139, line 650: Maybe we can revert directly if `from == to`, but it's ok, it's on your decide.

zhoujia6139 (approved):

PR #46, opened 2022-09-27
New fix and refactor

zhoujia6139:
## Security Checklist

- [ ] 1. Re-Entrancy
- [ ] 3. Unexpected Ether
- [ ] 4. Delegatecall
- [ ] 5. Default Visibilities
- [ ] 6. Entropy Illusion

### ⚠️ NOTES ⚠️

### DIFF
--- a/contracts/dependencies/uniswap/IMulticall.sol
+++ b/contracts/dependencies/uniswap/IMulticall.sol
@@ -1,0 +1,13 @@
+// SPDX-License-Identifier: GPL-2.0-or-later
+pragma solidity >=0.7.5;
+pragma abicoder v2;
+
+/// @title Multicall interface
+/// @notice Enables calling multiple methods in a single call to the contract
+interface IMulticall {
+    /// @notice Call multiple functions in the current contract and return the data from all of them if they all succeed
+    /// @dev The `msg.value` should not be trusted for any method callable from multicall.
+    /// @param data The encoded function data for each of the calls to make to this contract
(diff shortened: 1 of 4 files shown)


contracts/dependencies/uniswap/IMulticall.sol
```diff
@@ -0,0 +1,13 @@
...
+pragma abicoder v2;
+
+/// @title Multicall interface
+/// @notice Enables calling multiple methods in a single call to the contract
+interface IMulticall {
+    /// @notice Call multiple functions in the current contract and return the data from all of them if they all succeed
+    /// @dev The `msg.value` should not be trusted for any method callable from multicall.
+    /// @param data The encoded function data for each of the calls to make to this contract
+    /// @return results The results from each of the calls passed in via data
+    function multicall(bytes[] calldata data) external payable returns (bytes[] memory results);
```
GopherJ, line 12: how do we handle ETH? because msg.value check will always pass

zhoujia6139: no need. Uniswap handle it like this:
if msg.value is not enough, then revert the tx
if msg.value is more than needed, then refund

and this IMulticall is only used for unit test. not used in our contract logic.

GopherJ: ok I thought we plan to use it for `batchBuyWithCredit` etc

GopherJ (approved):

PR #48, opened 2022-09-27
feat: custom conduit

GopherJ:

## Security Checklist

- [ ] 1. Re-Entrancy
- [ ] 3. Unexpected Ether
- [ ] 4. Delegatecall
- [ ] 5. Default Visibilities
- [ ] 6. Entropy Illusion

### ⚠️ NOTES ⚠️


contracts/protocol/libraries/logic/MarketplaceLogic.sol
```diff
@@ -294,6 +294,16 @@ library MarketplaceLogic {
                 continue;
             }
 
+            if (
+                IERC721(token).ownerOf(tokenId) != onBehalfOf &&
+                INToken(reserve.xTokenAddress).ownerOf(tokenId) == onBehalfOf
+            ) {
+                if (!userConfig.isUsingAsCollateral(reserve.id)) {
+                    userConfig.setUsingAsCollateral(reserve.id, true);
```
WalidOfNow, line 302: is this enough after removing increasing the collaterlized balance for the receiver in this [PR](https://github.com/para-space/paraspace-core/pull/41)? we might need to set the token as collateral from the NToken right?

GopherJ: yes we need to set token as collateral, it's a bug

PR #55, opened 2022-10-01
chore: optimize supply erc721 logic

zhoujia6139:
## Security Checklist

- [ ] 1. Re-Entrancy
- [ ] 3. Unexpected Ether
- [ ] 4. Delegatecall
- [ ] 5. Default Visibilities
- [ ] 6. Entropy Illusion

### ⚠️ NOTES ⚠️

### DIFF
--- a/contracts/protocol/libraries/logic/SupplyLogic.sol
+++ b/contracts/protocol/libraries/logic/SupplyLogic.sol
@@ -131,21 +131,20 @@ library SupplyLogic {
     }
 
     function executeSupplyERC721Base(
-        mapping(address => DataTypes.ReserveData) storage reservesData,
+        uint16 reserveId,
+        address nTokenAddress,
         DataTypes.UserConfigurationMap storage userConfig,
         DataTypes.ExecuteSupplyERC721Params memory params
     ) internal {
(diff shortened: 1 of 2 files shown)


contracts/protocol/libraries/logic/SupplyLogic.sol
```diff
@@ -167,21 +166,28 @@ library SupplyLogic {
         DataTypes.UserConfigurationMap storage userConfig,
         DataTypes.ExecuteSupplyERC721Params memory params
     ) external {
+        DataTypes.ReserveData storage reserve = reservesData[params.asset];
+        DataTypes.ReserveCache memory reserveCache = reserve.cache();
```
WalidOfNow, line 170: we acgtually dont need to perform the .cache() operation since its too expensive. we can just pass the parameters that are needed. I think we only need the configuration, ntokenAddress and reserveId?

zhoujia6139: yeah. it's ture.

zhoujia6139:
@WalidOfNow  @GopherJ I just update the cache logic to reduce gas consumption, you guys can have a look.

WalidOfNow (approved):

PR #66, opened 2022-10-05
feat: finalize batch end auction

GopherJ:

## Security Checklist

- [ ] 1. Re-Entrancy
- [ ] 3. Unexpected Ether
- [ ] 4. Delegatecall
- [ ] 5. Default Visibilities
- [ ] 6. Entropy Illusion

### ⚠️ NOTES ⚠️


contracts/protocol/pool/PoolStorage.sol
```diff
@@ -31,4 +31,7 @@ contract PoolStorage {
 
     // Maximum allowed number of atomic tokens per user
     uint24 internal _maxAtomicTokensAllowed;
+
+    // Auction recovery health factor
+    uint256 internal _auctionRecoveryHealthFactor;
```
WalidOfNow, line 36: Does this require unit256? Or can we stack it with the previous state var?

contracts/protocol/pool/PoolParameters.sol
```diff
@@ -261,4 +248,37 @@ contract PoolParameters is
...
+        override
+        onlyPoolConfigurator
+    {
+        require(value != 0, Errors.INVALID_AMOUNT);
+
+        _auctionRecoveryHealthFactor = value;
+    }
+
+    /// @inheritdoc IPoolParameters
+    function setERC721HFVerifyTime(address user) external virtual override {
```
WalidOfNow, line 265: maybe we should add no-reentrancy guard on this as well

contracts/protocol/pool/PoolStorage.sol
WalidOfNow: also this is. global number for all reserves. are we sure we want to keep it this way or make it more dynamic per-reseerve?

WalidOfNow (approved):

PR #71, opened 2022-10-08
Fix/remove buy on behalf of

GopherJ:
## Security Checklist

- [ ] 1. Re-Entrancy
- [ ] 3. Unexpected Ether
- [ ] 4. Delegatecall
- [ ] 5. Default Visibilities
- [ ] 6. Entropy Illusion

### ⚠️ NOTES ⚠️

### DIFF
--- a/contracts/protocol/pool/PoolMarketplace.sol
+++ b/contracts/protocol/pool/PoolMarketplace.sol
@@ -143,6 +143,15 @@ contract PoolMarketplace is
             ).getAskOrderInfo(payload, WETH);
             orderInfo.taker = msg.sender;
 
+            // Once we encounter a listing using WETH, then we convert all our ethLeft to WETH
+            // this also means that the order is important
+            //
+            // the following example image that the `taker` owns only ETH and wants to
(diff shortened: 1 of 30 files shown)


contracts/protocol/pool/PoolMarketplace.sol
```diff
@@ -122,21 +135,43 @@ contract PoolMarketplace is
...
                 marketplace.adapter
-            ).getAskOrderInfo(payload, WETH);
-            orderInfo.taker = onBehalfOf;
+            ).getAskOrderInfo(payload, weth);
+            orderInfo.taker = msg.sender;
+
+            // Once we encounter a listing using WETH, then we convert all our ethLeft to WETH
+            // this also means that the parameters order is very important
+            //
+            // frontend/sdk needs to guarantee that WETH orders will always be put after ALL
```
zhoujia6139, line 144: how about add an explicit check if WETH orders have been put after ETH orders?

GopherJ: feel it requires looping through the orders

PR #73, opened 2022-10-09
refactor: liquidation logic

yrong:
## Security Checklist

- [ ] 1. Re-Entrancy
- [ ] 3. Unexpected Ether
- [ ] 4. Delegatecall
- [ ] 5. Default Visibilities
- [ ] 6. Entropy Illusion

### ⚠️ NOTES ⚠️

### DIFF
--- a/contracts/protocol/libraries/logic/LiquidationLogic.sol
+++ b/contracts/protocol/libraries/logic/LiquidationLogic.sol
@@ -86,7 +86,7 @@ library LiquidationLogic {
         uint256 userGlobalTotalDebt;
         uint256 userTotalDebt;
         uint256 actualDebtToLiquidate;
-        uint256 collateralDiscountedPrice;
+        uint256 actualLiquidationAmount;
         uint256 actualCollateralToLiquidate;
         uint256 liquidationBonus;
         uint256 healthFactor;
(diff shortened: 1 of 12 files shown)


yrong:
mainly follow items in improve list  
    ```
1. liquidationAsset vs debtAsset, liquidationAssetReserve vs debtReserve
        a. if we really want to allow non-borrowed asset for liquidation, we should always use `liquidationAsset` instead of debtAsset OR debtReserve
2. baseCollateral naming is very confusing, we can remove it and keep only `maxCollateralToLiquidate`
3. collateralAmount naming is very confusing, can it be renamed as `actualCollateralToLiquidate`
4. collateralDiscountedPrice is very confusing, can it be renamed as
    
    `actualLiquidationAmount`
    
5. erc721 liquidation shouldn’t read decimals ⇒ to keep consistency because other places we didn’t use
6. doesn’t respect CEI(check effects interaction)
```

for item 1 not change yet , for me both are ok  but prefer `debtAsset` from Aave a little bit more easy to understand than `liquidationAsset` , but I'm ok if we all agree to change @WalidOfNow @GopherJ

**Comment by GopherJ** [comment]
@yrong the issue is that we support liquidation using non-borrowed assets

File: contracts/protocol/libraries/logic/LiquidationLogic.sol [hunk 1]
```diff
@@ -344,6 +339,12 @@ library LiquidationLogic {
             vars.actualDebtToLiquidate
         );
 
+        if (params.receiveXToken) {
```
zhoujia6139, line 342: can we also move burn debtToken logic here.

```
        if (vars.userTotalDebt == vars.actualDebtToLiquidate) {
            userConfig.setBorrowing(debtReserve.id, false);
        }
        _burnDebtTokens(params, vars);
```

yrong:
> @yrong the issue is that we support liquidation using non-borrowed assets

@GopherJ so you mean in some cases debt asset to be in liquidation may not borrowed directly by the user, so better to use `liquidationAsset` to replace, right?

WalidOfNow (approved):

GopherJ (approved):

PR #74, opened 2022-10-09
Reduce ntoken size

zhoujia6139:
## Security Checklist

- [ ] 1. Re-Entrancy
- [ ] 3. Unexpected Ether
- [ ] 4. Delegatecall
- [ ] 5. Default Visibilities
- [ ] 6. Entropy Illusion

### ⚠️ NOTES ⚠️


test-suites/ntoken-punk.spec.ts
```diff
@@ -324,61 +324,6 @@ makeSuite("Punk nToken Mint and Burn Event Accounting", (testEnv) => {
...
-
-    const nWPunkBalance = await nWPunk.balanceOf(user3.address);
-    expect(nWPunkBalance).to.be.equal(1);
-
-    // signs signature
-    const chainId = hre.network.config.chainId || HARDHAT_CHAINID;
-    const deadline = MAX_UINT_AMOUNT;
-    const nonce = (await nWPunk.nonces(user3.address)).toNumber();
-    // use nft token length here as amount
-    const permitAmount = "1";
```
yrong, line 350: @zhoujia6139  `permitAmount` we sign here actually is length of the nft array which verified later in  `WPunkGateway` so should be fine I think, but anyway we can just remove it for safety

zhoujia6139: we can't use `punkIndexes` length as a token id. Image i call `withdrawPunkWithPermit` with `[4, 5, 6]` as `punkIndexes`, but only 3 is approved, so the transaction will fail.

yrong: https://github.com/para-space/paraspace-core/blob/270228a12435bc3056490ee72e910c18cbfe5b5a/contracts/ui/WPunkGateway.sol#L222

what I mean is that what we check in the contract is also length but not id.  but I'm with you that for ERC721 permit there should be some more secure way instead of use ERC20's permit. So remove it totally should be fine

yrong: yes, we need to pass token id into permit to make it approved.

GopherJ (approved):

PR #77, opened 2022-10-11
Use loadFixtures

alannotnerd:
- refactor: remove **setup**
- refactor: use fixtures rather than **setup**

## Security Checklist

- [x] 1. Re-Entrancy
- [x] 2. Arithmetic Over/Under Flows
- [x] 3. Unexpected Ether
- [x] 4. Delegatecall
- [x] 5. Default Visibilities
- [x] 6. Entropy Illusion
- [x] 7. External Contract Referencing
- [x] 8. Short Address/Parameter Attack (off chain)
- [x] 9. Unchecked CALL Return Values
- [x] 10. Race Conditions / Front Running
- [x] 11. Denial Of Service (DOS)
- [x] 12. Block Timestamp Manipulation
- [x] 13. Constructors with Care
- [x] 14. Uninitialized Storage Pointers
- [x] 15. Floating Points and Precision
- [x] 16. Tx.Origin Authentication
- [x] 17. Address.isContract Re-Entrancy via Constructor

### ⚠️ NOTES ⚠️


ivansolomonoff:
is this safe to run in parallel already? if so what would be the usage? we would like `make test` to parallelise runs

alannotnerd:
We want to make it able to be in parallel. But the deployment scripts have some issues. My concerns are: 
 - The `__setup__` is NOT special for mocha. What make the tests run well is all because of the alphabet order of `_`. I don't think it's a good practice.
 - We made a utility function called `makeSuite` to manage the EVM state manually, which is provided by the hardhat framework. It's also not elegant.

GopherJ (approved):

PR #83, opened 2022-10-13
refactor: remove stable rate model

yrong:
## Security Checklist

- [ ] 1. Re-Entrancy
- [ ] 3. Unexpected Ether
- [ ] 4. Delegatecall
- [ ] 5. Default Visibilities
- [ ] 6. Entropy Illusion

### ⚠️ NOTES ⚠️


contracts/protocol/libraries/logic/LiquidationLogic.sol
```diff
@@ -767,11 +757,13 @@ library LiquidationLogic {
             uint256
         )
     {
-        (uint256 userStableDebt, uint256 userVariableDebt) = Helpers
-            .getUserCurrentDebt(params.user, liquidationAssetReserveCache);
+        uint256 userVariableDebt = Helpers.getUserCurrentDebt(
```
ivansolomonoff, line 760: let's better call this userTotalDebt right away

contracts/protocol/libraries/logic/ValidationLogic.sol
```diff
@@ -424,132 +368,19 @@ library ValidationLogic {
...
-        ).getUserLastUpdated(onBehalfOf);
-
         require(
-            (stableRatePreviousTimestamp < uint40(block.timestamp) &&
-                interestRateMode == DataTypes.InterestRateMode.STABLE) ||
-                (variableDebtPreviousIndex <
-                    reserveCache.nextVariableBorrowIndex &&
-                    interestRateMode == DataTypes.InterestRateMode.VARIABLE),
+            (variableDebtPreviousIndex < reserveCache.nextVariableBorrowIndex &&
+                interestRateMode == DataTypes.InterestRateMode.VARIABLE),
```
ivansolomonoff, line 373: let's split this into 2 requires with the proper error, we have ProtocolError INVALID_INTEREST_RATE_MODE_SELECTED

contracts/protocol/libraries/logic/ValidationLogic.sol
```diff
@@ -424,132 +368,19 @@ library ValidationLogic {
...
             Errors.SAME_BLOCK_BORROW_REPAY
         );
 
         require(
-            (stableDebt != 0 &&
-                interestRateMode == DataTypes.InterestRateMode.STABLE) ||
-                (variableDebt != 0 &&
-                    interestRateMode == DataTypes.InterestRateMode.VARIABLE),
+            (variableDebt != 0 &&
+                interestRateMode == DataTypes.InterestRateMode.VARIABLE),
```
ivansolomonoff, line 379: and here you won't need to check for rate mode again

test-suites/helpers/validated-steps.ts
```diff
@@ -1056,15 +1058,16 @@ const liquidateAndValidateERC721 = async (
...
   // if isNFT and liquidationAsset is not being borrowed by the user,
   // then there's no bonus and liquidated amount supplied on behalf of the borrower,
   // so total collateral should remain the same)
-  if (!isLiquidationAssetBorrowed) {
-    assertAlmostEqual(
-      totalCollateral,
-      totalCollateralBefore.add(auctionExcessFunds)
-    );
-  } else {
+  // todo: seems none-borrow liquidation check will fail
```
ivansolomonoff, line 1068: does this really broke after removing stable rate?

contracts/protocol/libraries/logic/ValidationLogic.sol
yrong: since we remove stable rate, I would prefer also remove `InterestRateMode` totally to make less confusing, but will introduce some api break change, but I think should be fine before we release the production @ivansolomonoff @GopherJ  what do you think?

contracts/interfaces/IPoolCore.sol
```diff
@@ -69,7 +69,7 @@ interface IPoolCore {
      * initiator of the transaction on flashLoan()
      * @param onBehalfOf The address that will be getting the debt
      * @param amount The amount borrowed out
-     * @param interestRateMode The rate mode: 1 for Stable, 2 for Variable
+     * @param interestRateMode The rate mode
```
zhoujia6139, line 72: why don't we remove this `interestRateMode` directly.

contracts/protocol/libraries/logic/LiquidationLogic.sol
```diff
@@ -709,7 +704,7 @@ library LiquidationLogic {
         DataTypes.ExecuteLiquidationCallParams memory params,
         LiquidationCallLocalVars memory vars
     ) internal {
-        if (vars.userVariableDebt >= vars.actualDebtToLiquidate) {
+        if (vars.userTotalDebt >= vars.actualDebtToLiquidate) {
```
zhoujia6139, line 707: i think we don't need this if else logic anymore.

contracts/protocol/pool/DefaultReserveInterestRateStrategy.sol
```diff
@@ -252,79 +155,48 @@ contract DefaultReserveInterestRateStrategy is IReserveInterestRateStrategy {
...
     /**
      * @dev Calculates the overall borrow rate as the weighted average between the total variable debt and total stable
      * debt
-     * @param totalStableDebt The total borrowed from the reserve at a stable rate
      * @param totalVariableDebt The total borrowed from the reserve at a variable rate
      * @param currentVariableBorrowRate The current variable borrow rate of the reserve
-     * @param currentAverageStableBorrowRate The current weighted average of all the stable rate loans
      * @return The weighted averaged borrow rate
      **/
     function _getOverallBorrowRate(
```
zhoujia6139, line 184: we can also remove it now.

test-suites/helpers/validated-steps.ts
yrong: yes, tweak a little bit here to make test pass but seems nothing to do with this refactor.  I'm not sure if non-borrowed liquidation really work as expected but to make this pr focused I may trace it in other pr.

GopherJ (approved):

PR #96, opened 2022-10-17
separate non-borrowed liquidation & refactor validation step

yrong:
## Security Checklist

- [ ] 1. Re-Entrancy
- [ ] 3. Unexpected Ether
- [ ] 4. Delegatecall
- [ ] 5. Default Visibilities
- [ ] 6. Entropy Illusion

### ⚠️ NOTES ⚠️


yrong:
@ivansolomonoff 
 did some refactor work in this pr to wrap all params before and after liquidation into a result structure. I would suggest we do checks in one validation function instead of here and there in multiple places. 

I feel the `***validate***` function is still somewhat too long. But since we're going to rewrite/refactor all test cases into new structure I maynot touch it in this PR and we may improve it later if necessary.

GopherJ (approved):

ivansolomonoff:
> @ivansolomonoff @alannotnerd did some refactor work in this pr to wrap all params before and after liquidation into a result structure. I would suggest we do checks in one validation function instead of here and there in multiple places.
> 
> I feel the `***validate***` function is still somewhat too long. But since we're going to rewrite/refactor all test cases into new structure I maynot touch it in this PR and we may improve it later if necessary.

Ok, let's merge this one, and I'll continue the refactoring and recheck those validations.

ivansolomonoff (approved):

PR #109, opened 2022-10-20
Id xtoken

zhoujia6139:
## Security Checklist

- [ ] 1. Re-Entrancy
- [ ] 3. Unexpected Ether
- [ ] 4. Delegatecall
- [ ] 5. Default Visibilities
- [ ] 6. Entropy Illusion

### ⚠️ NOTES ⚠️


contracts/protocol/tokenization/NTokenMoonBirds.sol
```diff
@@ -32,6 +32,15 @@ contract NTokenMoonBirds is NToken, IMoonBirdBase {
         // Intentionally left blank
     }
 
+    function getNTokenIdentifier()
+        external
+        pure
+        override
+        returns (string memory)
+    {
+        return "NTokenMoonBirds";
```
yrong, line 41: is identifier already set when we `initReserves` ?

https://github.com/para-space/paraspace-deploy/blob/07b8cb8a3b12b1dd8af0a51fe15f58a5d171ef51/helpers/init-helpers.ts#L319-L323

contracts/protocol/libraries/logic/ValidationLogic.sol
```diff
@@ -835,11 +837,16 @@ library ValidationLogic {
             Errors.ZERO_ADDRESS_NOT_VALID
         );
 
+        INToken nToken = INToken(reserve.xTokenAddress);
+        require(
+            nToken.getXTokenType() == XTokenType.NTokenUniswapV3Type,
```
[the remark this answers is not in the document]
zhoujia6139: yeah. it's true. I'll commit my test case later.

contracts/ui/WETHGateway.sol
```diff
@@ -97,7 +97,7 @@ contract WETHGateway is ReentrancyGuard, IWETHGateway, OwnableUpgradeable {
     {
         uint256 variableDebt = Helpers.getUserCurrentDebt(
             onBehalfOf,
-            IPool(pool).getReserveData(address(WETH)).xTokenAddress
+            IPool(pool).getReserveData(address(WETH)).variableDebtTokenAddress
```
GopherJ, line 100: oh:)

contracts/interfaces/IXTokenType.sol
```diff
@@ -0,0 +1,26 @@
+// SPDX-License-Identifier: AGPL-3.0
+pragma solidity 0.8.10;
+
+/**
+ * @title IXTokenType
+ * @author ParallelFi
+ * @notice Defines the basic interface for an IXTokenType.
+ **/
+enum XTokenType {
+    NoneType, // unused
```
GopherJ, line 10: why do we need `NoneType`?

zhoujia6139: no special reason. common way to prevent uninitialize zero value.

GopherJ: why our AssetType doesn't need it?

contracts/protocol/tokenization/NTokenMoonBirds.sol
```diff
@@ -32,11 +33,15 @@ contract NTokenMoonBirds is NToken, IMoonBirdBase {
         // Intentionally left blank
     }
 
+    function getXTokenType() external pure override returns (XTokenType) {
```
GopherJ, line 36: I think we can simplify it as `XTokenType()` because `name` and `symbol` doesn't have `get` prefix

GopherJ:
could you also add `NonReentrancyGuard` for https://github.com/para-space/paraspace-core/blob/177486bc040369a55a5f98f1e360682778efee24/contracts/protocol/tokenization/NTokenUniswapV3.sol#L128

contracts/interfaces/IXTokenType.sol
zhoujia6139: I add this NoneType  with no special reason, so don't need to care about it.

contracts/protocol/tokenization/NTokenMoonBirds.sol
zhoujia6139: it's ok. we also have many get function style.

zhoujia6139:
> for

It's already added.

GopherJ (approved):

PR #111, opened 2022-10-20
feat: sushi as fallback oracle

yrong:
## Security Checklist

- [ ] 1. Re-Entrancy
- [ ] 3. Unexpected Ether
- [ ] 4. Delegatecall
- [ ] 5. Default Visibilities
- [ ] 6. Entropy Illusion

### ⚠️ NOTES ⚠️

### DIFF
--- a/contracts/misc/ParaSpaceFallbackOracle.sol
+++ b/contracts/misc/ParaSpaceFallbackOracle.sol
@@ -3,7 +3,7 @@ pragma solidity 0.8.10;
 
 import {INFTOracle} from "./interfaces/INFTOracle.sol";
 import {IUniswapV2Factory} from "./interfaces/IUniswapV2Factory.sol";
-import {IUniswapV2Router01} from "./interfaces/IUniswapV2Router01.sol";
+import {IUniswapV2Router} from "./interfaces/IUniswapV2Router.sol";
 import {IUniswapV2Pair} from "./interfaces/IUniswapV2Pair.sol";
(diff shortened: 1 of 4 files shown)


yrong:
uniswap/sushiswap only as fallback should be fine but not sure if used as primary oracle source, or use some offchain twap oracle?

GopherJ:
@yrong we may need to use uniswap v3 twap oracle

yrong:
> @yrong we may need to use uniswap v3 twap oracle if you are interested

 actually this pr is for fetch price from sushiswap and seems it does not support twap on chain

jfzhou5:
For security reasons, we should still have an off-chain oracle with TWAP.

contracts/misc/ParaSpaceFallbackOracle.sol
```diff
@@ -40,24 +46,36 @@ contract ParaSpaceFallbackOracle {
...
         address pairAddress = IUniswapV2Factory(UNISWAP_FACTORY).getPair(
             WETH,
             asset
         );
+        if (pairAddress == address(0x00)) {
+            pairAddress = IUniswapV2Factory(SUSHISWAP_FACTORY).getPair(
+                WETH,
+                asset
+            );
+            if (pairAddress != address(0x00)) {
```
zhoujia6139, line 59: don't need this check?
```
if (pairAddress != address(0x00))  {
}
```

yrong:
as discussed, single source price from swap may not be safe and will not do it in this way.

PR #112, opened 2022-10-20
feat: always use pool as NFT purchase recipient

GopherJ:

## Security Checklist

- [x] 1. Re-Entrancy
- [x] 2. Arithmetic Over/Under Flows
- [x] 3. Unexpected Ether
- [x] 4. Delegatecall
- [x] 5. Default Visibilities
- [x] 6. Entropy Illusion
- [x] 7. External Contract Referencing
- [x] 8. Short Address/Parameter Attack (off chain)
- [x] 9. Unchecked CALL Return Values
- [x] 10. Race Conditions / Front Running
- [x] 11. Denial Of Service (DOS)
- [x] 12. Block Timestamp Manipulation
- [x] 13. Constructors with Care
- [x] 14. Uninitialized Storage Pointers
- [x] 15. Floating Points and Precision
- [x] 16. Tx.Origin Authentication
- [x] 17. Address.isContract Re-Entrancy via Constructor

### ⚠️ NOTES ⚠️


contracts/protocol/libraries/logic/MarketplaceLogic.sol
```diff
@@ -499,10 +497,7 @@ library MarketplaceLogic {
                 bool isNToken = reservesData[underlyingAsset].xTokenAddress ==
```
zhoujia6139, line 497: If front end always buy or bid the underlying asset, then we can remove this branch?
```
if (vars.xTokenAddress == address(0)) {
}
```

GopherJ: for moonbirds I'll ask them

contracts/protocol/libraries/logic/MarketplaceLogic.sol
```diff
@@ -499,10 +497,7 @@ library MarketplaceLogic {
...
                     token;
                 require(isNToken, Errors.ASSET_NOT_LISTED);
-                require(
-                    INToken(token).ownerOf(tokenId) == onBehalfOf,
-                    Errors.INVALID_RECIPIENT
-                );
+                IERC721(token).transferFrom(address(this), onBehalfOf, tokenId);
```
yrong, line 500: use `safeTransferFrom` instead?

contracts/protocol/libraries/logic/MarketplaceLogic.sol
```diff
@@ -516,7 +511,12 @@ library MarketplaceLogic {
 
             // item.token == underlyingAsset but supplied after listing/offering
             // so NToken is transferred instead
-            if (INToken(vars.xTokenAddress).ownerOf(tokenId) == onBehalfOf) {
+            if (INToken(vars.xTokenAddress).ownerOf(tokenId) == address(this)) {
```
yrong, line 514: if from accept bid seems we can still keep this `onBehalfOf` short circuit?
```
if (
                INToken(vars.xTokenAddress).ownerOf(tokenId) == address(this) ||
                INToken(vars.xTokenAddress).ownerOf(tokenId) == onBehalfOf
            )
```

contracts/protocol/libraries/logic/MarketplaceLogic.sol
```diff
@@ -516,7 +511,12 @@ library MarketplaceLogic {
...
+                IERC721(vars.xTokenAddress).transferFrom(
```
yrong, line 515: `safeTransferFrom`

contracts/protocol/libraries/logic/MarketplaceLogic.sol
GopherJ: for NToken it's the same because we didn't implement the real `safeTransferFrom` yet

GopherJ: I can switch to that in case  one day we change `safeTransferFrom` to add back the hook

contracts/protocol/libraries/logic/MarketplaceLogic.sol
GopherJ: no matter if received underlying NFT or NToken it'll fail

PR #122, opened 2022-10-25
WIP: support upgrade xtoken

zhoujia6139:
## Security Checklist

- [ ] 1. Re-Entrancy
- [ ] 3. Unexpected Ether
- [ ] 4. Delegatecall
- [ ] 5. Default Visibilities
- [ ] 6. Entropy Illusion

### ⚠️ NOTES ⚠️

### DIFF
--- a/contracts/protocol/libraries/logic/ConfiguratorLogic.sol
+++ b/contracts/protocol/libraries/logic/ConfiguratorLogic.sol
@@ -27,7 +27,12 @@ library ConfiguratorLogic {
         address interestRateStrategyAddress,
         address auctionStrategyAddress
     );
-    event XTokenUpgraded(
+    event PTokenUpgraded(
+        address indexed asset,
+        address indexed proxy,
+        address indexed implementation
+    );
+    event NTokenUpgraded(
         address indexed asset,
         address indexed proxy,
         address indexed implementation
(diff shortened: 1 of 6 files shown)


contracts/protocol/libraries/logic/ConfiguratorLogic.sol
```diff
@@ -157,7 +162,44 @@ library ConfiguratorLogic {
             encodedCall
```
yrong, line 162: seems we wrap encodedCall as `initialize` and call it again, is that necessary for upgrade or not?

contracts/mocks/upgradeability/MockNToken.sol
```diff
@@ -0,0 +1,14 @@
...
+
+import {NToken} from "../../protocol/tokenization/NToken.sol";
+import {IPool} from "../../interfaces/IPool.sol";
+import {IRewardController} from "../../interfaces/IRewardController.sol";
+
+contract MockNToken is NToken {
+    constructor(IPool pool) NToken(pool, false) {}
+
+    function getRevision() internal pure override returns (uint256) {
+        return 0x2;
```
yrong, line 12: seems run `initialize` is required to update revision here

PR #135, opened 2022-10-29
Fix: Ensure there is liquidity to release borrowed tokens

ivansolomonoff:
## Security Checklist

- [ ] 1. Re-Entrancy
- [ ] 3. Unexpected Ether
- [ ] 4. Delegatecall
- [ ] 5. Default Visibilities
- [ ] 6. Entropy Illusion

### ⚠️ NOTES ⚠️


contracts/protocol/libraries/logic/ValidationLogic.sol
```diff
@@ -252,6 +252,17 @@ library ValidationLogic {
             Errors.PRICE_ORACLE_SENTINEL_CHECK_FAILED
         );
 
+        if (params.releaseUnderlying) {
+            vars.availableLiquidity = IERC20(params.asset).balanceOf(
+                params.reserveCache.xTokenAddress
```
GopherJ, line 257: should take `reserves` into account

contracts/protocol/libraries/logic/ValidationLogic.sol
```diff
@@ -252,6 +252,17 @@ library ValidationLogic {
...
+            );
+
+            require(
```
yrong, line 260: if we already check collateral can cover debt [here](https://github.com/para-space/paraspace-core/blob/da9d57291c01f33de106819347a624f4ae84b181/contracts/protocol/libraries/logic/ValidationLogic.sol#L327-L331) do we still need this?

contracts/protocol/libraries/logic/ValidationLogic.sol
```diff
@@ -252,6 +252,17 @@ library ValidationLogic {
...
+                vars.availableLiquidity >= params.amount,
+                Errors.NOT_ENOUGH_LIQUIDITY_IN_RESERVE
+            );
```
yrong, line 263: if we already check collateral can always cover debt [here](https://github.com/para-space/paraspace-core/blob/da9d57291c01f33de106819347a624f4ae84b181/contracts/protocol/libraries/logic/ValidationLogic.sol#L327-L331) do we still need this?

contracts/protocol/libraries/logic/ValidationLogic.sol
zhoujia6139: I think there is no need to.

contracts/protocol/libraries/logic/ValidationLogic.sol
zhoujia6139: Yes, we need. what we are checking here is if cash is enough for borrowing, not LTV.

ivansolomonoff: yeah, basically we want to make sure that when the pool is releasing tokens to the users, that it actually has them.

WalidOfNow:
I personally don't think this is necessary as it adds gas cost to each tx. it can be helpful at the initial stage but then it just becomes an unneeded gas cost.

GopherJ:
@ivansolomonoff talked with the team and we agree with walid that this may not be necessary

PR #138, opened 2022-10-30
feat: close factor & collateral swap & liquidation refactorings

GopherJ:

## Security Checklist

- [ ] 1. Re-Entrancy
- [ ] 3. Unexpected Ether
- [ ] 4. Delegatecall
- [ ] 5. Default Visibilities
- [ ] 6. Entropy Illusion

### ⚠️ NOTES ⚠️


contracts/protocol/libraries/logic/LiquidationLogic.sol
```diff
@@ -602,21 +633,34 @@ library LiquidationLogic {
...
 
-        uint256 actualDebtToLiquidate = params.liquidationAmount > userTotalDebt
-            ? userTotalDebt
+        uint256 closeFactor = healthFactor > CLOSE_FACTOR_HF_THRESHOLD
+            ? DEFAULT_LIQUIDATION_CLOSE_FACTOR
+            : MAX_LIQUIDATION_CLOSE_FACTOR;
+
+        uint256 maxLiquidatableDebt = userTotalDebt.percentMul(
+            closeFactor > minimumCloseFactor ? closeFactor : minimumCloseFactor
+        );
```
yrong, line 659: will this apply to nft or only erc20?

zhoujia6139: i think should only for erc20.

contracts/protocol/libraries/logic/LiquidationLogic.sol
```diff
@@ -142,81 +185,64 @@ library LiquidationLogic {
...
-                vars.liquidationProtocolFeeAmount
-            );
-        }
-
         // If the collateral being liquidated is equal to the user balance,
         // we set the currency as not being used as collateral anymore
-        if (vars.actualCollateralToLiquidate == vars.userCollateralBalance) {
+        if (
+            vars.actualCollateralToLiquidate + vars.liquidationProtocolFee ==
+            vars.userCollateral
```
yrong, line 227: will it be safer to use `>=` instead?

contracts/protocol/libraries/logic/LiquidationLogic.sol
```diff
@@ -256,22 +282,22 @@ library LiquidationLogic {
...
             params.liquidationAsset
         ];
         DataTypes.UserConfigurationMap storage userConfig = usersConfig[
             params.user
         ];
-        uint16 liquidationAssetReserveId = liquidationAssetReserve.id;
+
+        vars.liquidationAssetReserveId = liquidationAssetReserve.id;
         vars.liquidationAssetReserveCache = liquidationAssetReserve.cache();
+        liquidationAssetReserve.updateState(vars.liquidationAssetReserveCache);
```
yrong, line 294: is `updateState` required here? seems already covered in supply logic

contracts/protocol/libraries/logic/LiquidationLogic.sol
```diff
@@ -959,41 +622,40 @@ library LiquidationLogic {
...
         DataTypes.ReserveData storage collateralReserve,
-        DataTypes.ReserveCache memory liquidationAssetReserveCache,
-        address collateralAsset,
-        address liquidationAsset,
-        uint256 liquidationAmount,
-        uint256 userCollateralBalance,
-        uint256 liquidationBonus,
-        IPriceOracleGetter oracle
+        DataTypes.ExecuteLiquidationCallParams memory params,
+        LiquidationCallLocalVars memory superVars
```
yrong, line 635: will it better to rename as `callVars`?

contracts/protocol/libraries/logic/ValidationLogic.sol
```diff
@@ -578,7 +578,7 @@ library ValidationLogic {
         }
 
         require(
-            params.liquidationAmount >= params.actualLiquidationAmount,
+            params.maxLiquidationAmount >= params.actualLiquidationAmount,
             Errors.LIQUIDATION_AMOUNT_NOT_ENOUGH
         );
```
yrong, line 584: https://github.com/para-space/paraspace-core/blob/379d9e923a578d0b8c17778fcbdbfe8141593b26/contracts/protocol/libraries/logic/ValidationLogic.sol#L597-L600

maybe better to add another error code for this check

contracts/protocol/libraries/logic/LiquidationLogic.sol
zhoujia6139: seems not need here.

contracts/protocol/libraries/logic/LiquidationLogic.sol
```diff
@@ -142,81 +185,64 @@ library LiquidationLogic {
                 })
             );
 
-        (vars.userTotalDebt, vars.actualDebtToLiquidate) = _calculateDebt(
-            vars.liquidationAssetReserveCache,
-            params
+        (vars.userDebt, vars.actualDebtToLiquidate) = _calculateDebt(
```
WalidOfNow, line 188: why are we refactoring the ERC20 liquidation? aave's code has been audited and has been working on mainnet without issues why can't we keep it as is and just add back the close factor?

contracts/protocol/libraries/logic/LiquidationLogic.sol
```diff
@@ -38,6 +38,27 @@ library LiquidationLogic {
...
+     * Expressed in bps, a value of 1e4 results in 100.00%
+     */
+    uint256 public constant MAX_LIQUIDATION_CLOSE_FACTOR = 1e4;
+
+    /**
+     * @dev This constant represents below which health factor value it is possible to liquidate
+     * an amount of debt corresponding to `MAX_LIQUIDATION_CLOSE_FACTOR`.
+     * A value of 0.85e18 results in 85%
+     */
+    uint256 public constant CLOSE_FACTOR_HF_THRESHOLD = 0.85e18;
```
WalidOfNow, line 60: how did we decide on this number?

contracts/protocol/libraries/logic/LiquidationLogic.sol
```diff
@@ -367,375 +366,63 @@ library LiquidationLogic {
...
-    ) internal {
-        DataTypes.ReserveData storage liquidationAssetReserve = reservesData[
-            params.liquidationAsset
-        ];
         // Transfer fee to treasury if it is non-zero
-        if (vars.liquidationProtocolFeeAmount != 0) {
+        if (vars.liquidationProtocolFee != 0) {
             IERC20(params.liquidationAsset).safeTransferFrom(
-                vars.liquidator,
+                params.liquidator,
```
ivansolomonoff, line 415: Logic-wise, it seems inconsistent that in ERC-20 case, borrower pays for the protocol fees (extra discount on their collateral), whereas in ERC-721 case, liquidator pays for them. This makes the calculations for actualLiquidationAmount behave differently which might not be a good idea. My personal opinion is that we should stick to the concept that fees as taken from borrower's collateral in the form of an extra discount of the price paid for the NFT. So liquidationProtocolFee in both cases penalise the borrower but not the liquidator.

PR #155, opened 2022-11-03
Uniswapv3 check rule

zhoujia6139:
## Security Checklist

- [ ] 1. Re-Entrancy
- [ ] 3. Unexpected Ether
- [ ] 4. Delegatecall
- [ ] 5. Default Visibilities
- [ ] 6. Entropy Illusion

### ⚠️ NOTES ⚠️


contracts/protocol/pool/PoolCore.sol
```diff
@@ -253,6 +231,35 @@ contract PoolCore is
             );
     }
 
+    function decreaseUniswapV3Liquidity(
```
WalidOfNow, line 234: we should probably make a new Pool contract and call it PoolUniswapV3.sol and add all the uniswap selectors to it.

contracts/protocol/libraries/logic/SupplyLogic.sol
```diff
@@ -173,6 +174,19 @@ library SupplyLogic {
             DataTypes.AssetType.ERC721
         );
 
+        INToken nToken = INToken(reserveCache.xTokenAddress);
+        if (nToken.getXTokenType() == XTokenType.NTokenUniswapV3) {
```
WalidOfNow, line 178: this kind of checks doesn't seem like a scalable solution. I do prefer having a separate supply function for uniswapV3 comparing to this ad-hoc solution.

zhoujia6139: if so, then we have too much repeated code logic for supply, withdraw, setUserAsCollateral, finalizeTransfer and liquidate.
Most important, we don't have a strong scalability requirement right now, and if we do have in the future, we can change at that moment.

yz89: Even if univ3's interface is separated, we need additional checks to prevent users supplying univ3 NFTS using this interface

GopherJ (approved):

zhoujia6139:
supply:active  pause frozen  
withdraw: active pause
borrow: active  pause frozen
repay: active pause
setCollateral: active pause
liquidate: active pause
auction: active pause
transfer: pause
flashloan: active pause

uniswapv3:
supply: active  pause frozen  
withdraw: active pause
setCollateral: active pause
decreaseLiquidity: active pause
transfer:pause
liquidate: active pause

PR #207, opened 2022-11-12
refactor: nft oracle

yrong:
## Security Checklist

- [ ] 1. Re-Entrancy
- [ ] 3. Unexpected Ether
- [ ] 4. Delegatecall
- [ ] 5. Default Visibilities
- [ ] 6. Entropy Illusion

### ⚠️ NOTES ⚠️


yrong:
- [x] refactor aggregation logic to use price from as much oracles as possible
- [x] introduce tidy mode with only required fixture for unit test

test-suites/helpers/make-suite.ts
```diff
@@ -476,3 +476,182 @@ export async function initializeMakeSuite() {
...
+    variableDebtDai: {} as VariableDebtToken,
+    variableDebtWeth: {} as VariableDebtToken,
+    pUsdc: {} as PToken,
+    usdc: {} as MintableERC20,
+    usdt: {} as MintableERC20,
+    nBAYC: {} as NTokenBAYC,
+    nMOONBIRD: {} as NTokenMoonBirds,
+    bayc: {} as MintableERC721,
+    addressesProvider: {} as PoolAddressesProvider,
+    registry: {} as PoolAddressesProviderRegistry,
```
GopherJ, line 510: do we really need so many if we would like to have a seperate env for oracle

GopherJ: I think  for now it's better to keep simple, this can be done in another pr to have multiple envs

.env.example
```diff
@@ -12,3 +12,7 @@ DB_PATH=:memory: #required
 DEPLOY_START=0
 DEPLOY_END=19
 # FORK=mainnet
+TIDY_TEST=false
+TEST_ONLY=false
+NFT_ORACLE_PRICE_EXPIRATION=1800
```
GopherJ, line 17: nft config doesn't change so much, so if we really need to have config per network. We can put inside market-config

test-suites/helpers/make-suite.ts
yrong: @GopherJ I've removed the tidy mode for simplicity, for `test_only` I would suggest we keep it, in this case we launch a live hardhat network locally and run deploy only once, then just rerun each test without deploy anything again, helps  a lot when I try to refactor oracle tests

contracts/misc/ERC721OracleWrapper.sol
```diff
@@ -48,12 +51,24 @@ contract ERC721OracleWrapper is IEACAggregatorProxy {
...
     function decimals() external pure override returns (uint8) {
         return 18;
     }
 
     function latestAnswer() external view override returns (int256) {
-        return int256(uint256(oracleAddress.getTwap(asset)));
+        uint256 lastUpdatedTime = oracleAddress.getLastUpdateTime(asset);
+        int256 lastUpdateTwap = int256(oracleAddress.getTwap(asset));
+        if ((block.number - lastUpdatedTime) > expirationPeriod) {
+            lastUpdateTwap = 0;
```
GopherJ, line 69: revert is always better than returning 0

contracts/misc/NFTFloorOracle.sol
```diff
@@ -59,14 +52,9 @@ struct FeederRegistrar {
 contract NFTFloorOracle is Initializable, AccessControl, INFTFloorOracle {
     event AssetAdded(address indexed asset);
     event AssetRemoved(address indexed asset);
-    event SetAssetData(address indexed asset, uint128 price, uint64 timestamp);
+    event SetAssetData(address indexed asset, uint256 price, uint256 timestamp);
```
GopherJ, line 55: shall it be AssetDataSet to align with OracleConfigSet

contracts/misc/NFTFloorOracle.sol
```diff
@@ -235,34 +207,32 @@ contract NFTFloorOracle is Initializable, AccessControl, INFTFloorOracle {
...
             return true;
         }
-        if (_price > price) {
-            percentDeviation = _price / price;
-        } else {
-            percentDeviation = price / _price;
-        }
-        if (percentDeviation >= config.maxPriceDeviation) {
+        priceDeviation = _price > price
+            ? (_price * 100) / price
```
GopherJ, line 232: I think it's better to use `wadDiv`

contracts/misc/NFTFloorOracle.sol
```diff
@@ -182,34 +170,18 @@ contract NFTFloorOracle is Initializable, AccessControl, INFTFloorOracle {
         _setupRole(UPDATER_ROLE, admin);
```
GopherJ, line 170: admin may only need DEFAULT_ADMIN_ROLE because DEFAULT_ADMIN_ROLE hasRole(UPDATER_ROLE)

contracts/misc/NFTFloorOracle.sol
```diff
@@ -287,82 +257,79 @@ contract NFTFloorOracle is Initializable, AccessControl, INFTFloorOracle {
...
-    function addRawValue(address token, uint128 twap) internal {
+    function addRawValue(address token, uint256 twap) internal {
         FeederRegistrar storage feederRegistrar = priceFeederMap[token];
-        PricePriceInformationBuffer storage priceBuffer = feederRegistrar
+        PriceInformationBuffer storage priceBuffer = feederRegistrar
             .feederBuffer[msg.sender];
-        if (priceBuffer.next >= config.maxSubmissions) {
-            priceBuffer.next -= config.maxSubmissions;
+        if (priceBuffer.next >= MAX_SUBMISSION) {
+            priceBuffer.next -= MAX_SUBMISSION;
```
GopherJ, line 284: how about using module

contracts/misc/NFTFloorOracle.sol
yrong: for `DEFAULT_ADMIN_ROLE` we just accept price without execute any price check and aggregation logic, maybe useful in some  scenarios for price correction or emergency shutdown

contracts/misc/NFTFloorOracle.sol
yrong: maybe not better than deviation

contracts/misc/NFTFloorOracle.sol
yrong: ？

contracts/misc/NFTFloorOracle.sol
```diff
@@ -305,37 +293,27 @@ contract NFTFloorOracle is Initializable, AccessControl, INFTFloorOracle {
...
-                        validNum++;
-                    }
+        for (uint256 j = 0; j < feeders.length; j++) {
+            PriceInformation memory priceInfo = feederRegistrar.feederPrice[
+                feeders[j]
+            ];
+            if (priceInfo.lastUpdateTime > 0) {
+                uint256 diffTime = currentTime - priceInfo.lastUpdateTime;
+                if (diffTime < config.expirationPeriod) {
+                    validPriceList[validNum] = priceInfo.twap;
```
yz89, line 306: use validPriceList.push

contracts/misc/NFTFloorOracle.sol
```diff
@@ -305,37 +293,27 @@ contract NFTFloorOracle is Initializable, AccessControl, INFTFloorOracle {
...
-                PriceInformationBuffer memory feederBuffer = feederRegistrar
-                    .feederBuffer[feeders[j]];
-                PriceInformation memory priceInfo = feederBuffer.ring[i];
-                if (priceInfo.lastUpdateTime > 0) {
-                    uint256 diffTime = currentTime - priceInfo.lastUpdateTime;
-                    if (diffTime < config.expirationPeriod) {
-                        validPriceList[validNum] = priceInfo.twap;
-                        validNum++;
-                    }
+        for (uint256 j = 0; j < feeders.length; j++) {
```
yz89, line 299: use i at first

contracts/misc/NFTFloorOracle.sol
```diff
@@ -305,37 +293,27 @@ contract NFTFloorOracle is Initializable, AccessControl, INFTFloorOracle {
...
-                break;
-            }
         }
-
-        // only calculate median value when number of valid data greater than minCountToAggregate
-        if (validNum >= feeders.length) {
-            // ignore sort for saving gas, we just pick fixed element from validPriceList
-            return (true, validPriceList[validNum / 2]);
+        //we allow at most 1 feeder down
+        if (validNum < (feeders.length - 1)) {
```
yz89, line 312: use validPriceList.length

yrong: `validPriceList.length==feeders.length` we leave room for 1 oracle down here

test-suites/_oracle_nft_floor_price.spec.ts
```diff
@@ -332,51 +342,36 @@ describe("NFT Oracle Tests", () => {
...
+    //set price at block 7
     expect(
       await nftFloorOracle
         .connect(user1.signer)
-        .setPrice(mockToken.address, parseEther("5").toString())
+        .setPrice(mockToken.address, parseEther("4").toString())
     );
-    //and with user1:[1(expired),3,5] and user2:[2] so aggregated to [3,2]
-    //and 2 will be finalized
+    //price from user2 expired and with user1:4 and user3:5 so aggregated to [4,5]
```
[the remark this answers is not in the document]
yrong: https://github.com/para-space/paraspace-core/blob/08f0daeab19fb6a543ca5c588c5d5e7adaf450de/contracts/misc/NFTFloorOracle.sol#L368

use previous twap price

.env.example
```diff
@@ -12,3 +12,4 @@ DB_PATH=:memory: #required
 DEPLOY_START=0
 DEPLOY_END=19
 # FORK=mainnet
+TEST_ONLY=false
```
GopherJ, line 15: what's this for?

contracts/misc/ERC721OracleWrapper.sol
```diff
@@ -34,11 +35,13 @@ contract ERC721OracleWrapper is IEACAggregatorProxy {
     constructor(
         address _provider,
         address _oracleAddress,
-        address _asset
+        address _asset,
+        uint256 _expirationPeriod
```
GopherJ, line 39: I think it's hard to sync expirationPeriod between OracleWrapper & NFTFloorOracle, it's better to do it inside getTwap

GopherJ (commented):
I think we should simplify nft-oracle contracts, it has too many unecessary things and it didn't distinguish internal and external calls which makes it hard to review.

I think we can:

- add _ prefix for all internal calls
- make all storage variables public so that we dont necessary to add so many getters

contracts/misc/NFTFloorOracle.sol
yrong: change to 2/3 as threshold,  need to confirm with team @yz89

contracts/misc/NFTFloorOracle.sol
```diff
@@ -287,82 +260,79 @@ contract NFTFloorOracle is Initializable, AccessControl, INFTFloorOracle {
...
-                .feederBuffer[feeders[i]];
-            for (uint256 j = 0; j < priceBuffer.ring.length; j++) {
-                PriceInformation memory priceInfo = priceBuffer.ring[j];
+        //aggeregate with price in each ring position of all feeders
+        for (uint256 i = 0; i < MAX_SUBMISSION; i++) {
+            for (uint256 j = 0; j < feeders.length; j++) {
+                PriceInformationBuffer memory feederBuffer = feederRegistrar
+                    .feederBuffer[feeders[j]];
+                PriceInformation memory priceInfo = feederBuffer.ring[i];
                 if (priceInfo.lastUpdateTime > 0) {
```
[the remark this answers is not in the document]
yrong: @GopherJ  seems you are reviewing the outdated code, we refactor to now use ring buffer here in previous commit

contracts/misc/NFTFloorOracle.sol
```diff
@@ -287,82 +260,79 @@ contract NFTFloorOracle is Initializable, AccessControl, INFTFloorOracle {
...
-                    uint64 laggingTimestamp = _timestamp -
-                        priceInfo.lastUpdateTime;
-                    if (laggingTimestamp <= config.expirationPeriod) {
-                        //since it is memory array we can not push here
+                    uint256 diffTime = currentTime - priceInfo.lastUpdateTime;
+                    if (diffTime < config.expirationPeriod) {
```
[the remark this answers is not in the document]
yrong: same outdated

.env.example
yrong: https://github.com/para-space/paraspace-core/pull/207#discussion_r1022288561
@GopherJ commented here

contracts/misc/ERC721OracleWrapper.sol
yrong: set in OracleWrapper allow us to set different `expirationPeriod` for each asset, and we can set to same value when deploy and change it as we want later, it is a configurable param.

contracts/misc/NFTFloorOracle.sol
yrong: still a little confusing

contracts/misc/ERC721OracleWrapper.sol
GopherJ: @yrong but image that we changed `expirationPeriod` in NFTFloorOracle and we didn't update here. Will that cause issues?

yrong: we can prepare a script for change both values if necessary, and I do think separate to 2 config could be helpful. different use case one for global and one for each asset.

contracts/misc/NFTFloorOracle.sol
```diff
@@ -272,108 +238,89 @@ contract NFTFloorOracle is Initializable, AccessControl, INFTFloorOracle {
...
-            }
         }
-
-        // only calculate median value when number of valid data greater than minCountToAggregate
-        if (validNum >= config.minCountToAggregate) {
-            // ignore sort for saving gas, we just pick fixed element from validPriceList
-            // sort(validPriceList, 0, validNum);
-            return (true, validPriceList[config.minCountToAggregate / 2]);
+        //we will add 3-5 feeders at mainet launch and allow less than 1/3 feeders down
+        uint256 validNumThreshold = ((2 * (feeders.length)) / 3) < 3
```
zhoujia6139, line 308: should avoid calculate `((2 * (feeders.length)) / 3) `twice.

zhoujia6139:
For the `setOracles` function. can we replace with two function `addOracles` and `removeOracles`.

yrong:
> For the `setOracles` function. can we replace with two function `addOracles` and `removeOracles`.

then we need some extra logic or storage to maintain feeders and make sure length is always correct and without duplication, maybe we can improve it later if nessessary

contracts/misc/NFTFloorOracle.sol
yrong: use fixed value as threshold instead

yz89 (approved):

PR #232, opened 2022-11-19
Refactor/nft oracle

yrong:
## Security Checklist

- [ ] 1. Re-Entrancy
- [ ] 3. Unexpected Ether
- [ ] 4. Delegatecall
- [ ] 5. Default Visibilities
- [ ] 6. Entropy Illusion

### ⚠️ NOTES ⚠️


foundry.toml
```diff
@@ -0,0 +1,15 @@
+[profile.default]
+via_ir = false
+src = 'contracts'
+out = 'foundry_out'
+libs = ["foundry_libs"]
+test = 'foundry_test'
```
alannotnerd, line 6: Why use the prefix?

yrong: we may use multiple test frameworks to differ from one another

GopherJ (approved):
LGTM

PR #258, opened 2022-12-02
Ntoken bakc implementation

zhoujia6139:
## Security Checklist

- [ ] 1. Re-Entrancy
- [ ] 3. Unexpected Ether
- [ ] 4. Delegatecall
- [ ] 5. Default Visibilities
- [ ] 6. Entropy Illusion

### ⚠️ NOTES ⚠️


contracts/protocol/pool/PoolApeStaking.sol
```diff
@@ -120,20 +120,16 @@ contract PoolApeStaking is
...
-        {
-            DataTypes.ReserveData storage nftReserve = ps._reserves[nftAsset];
-            xTokenAddress = nftReserve.xTokenAddress;
-        }
-
-        INTokenApeStaking nTokenApeStaking = INTokenApeStaking(xTokenAddress);
-        IERC721 bakcContract = nTokenApeStaking.getBAKC();
+        address xTokenAddress = ps._reserves[nftAsset].xTokenAddress;
+        IERC721 bakcContract = INTokenApeStaking(xTokenAddress).getBAKC();
+        address bakcNToken = ps._reserves[address(bakcContract)].xTokenAddress;
```
GopherJ, line 130: bakcXTokenAddress and bakcToken are maybe better because they are token and address

contracts/protocol/pool/PoolApeStaking.sol
```diff
@@ -120,20 +120,16 @@ contract PoolApeStaking is
...
         uint256[] memory transferredTokenIds = new uint256[](_nftPairs.length);
+        address[] memory originArray = new address[](_nftPairs.length);
```
GopherJ, line 132: will `transferredTokenOwners` be more clear?

GopherJ (approved):

contracts/interfaces/IPoolApeStaking.sol
```diff
@@ -108,4 +108,21 @@ interface IPoolApeStaking {
...
         uint256[][] calldata tokenIds
     ) external;
+
+    /**
+     * @notice Claim user BAKC paired Ape coin reward and deposit to ape compound to get cApe, then deposit cApe to Lending pool for user
+     * @param nftAsset Contract address of BAYC/MAYC
+     * @param users array of user address
+     * @param _nftPairs Array of Paired BAYC/MAYC NFT's
+     */
+    function claimPairedApeRewardAndCompound(
```
GopherJ, line 118: claim already implies `Reward`, will `claimPairedApeAndCompound`  be simpler?

contracts/protocol/pool/PoolApeStaking.sol
```diff
@@ -489,6 +492,105 @@ contract PoolApeStaking is
...
+            ._reserves[address(localVar.bakcContract)]
+            .xTokenAddress;
+        localVar.balanceBefore = APE_COIN.balanceOf(address(this));
+        localVar.amounts = new uint256[](_nftPairs.length);
+
+        for (uint256 i = 0; i < _nftPairs.length; i++) {
+            ApeCoinStaking.PairNft[] calldata nftPair = _nftPairs[i];
+            localVar.originArray = new address[](nftPair.length);
+            for (uint256 j = 0; j < nftPair.length; j++) {
+                localVar.tmpPositionOwner = INToken(localVar.xTokenAddress)
```
GopherJ, line 530: why do we need to store this in `localVar`?

contracts/protocol/pool/PoolApeStaking.sol
```diff
@@ -489,6 +492,105 @@ contract PoolApeStaking is
...
+        uint256 totalFee = localVar.totalAmount.percentMul(compoundFee);
+        APE_COMPOUND.deposit(address(this), localVar.totalAmount);
+
+        if (totalFee > 0) {
+            IERC20(address(APE_COMPOUND)).safeTransfer(msg.sender, totalFee);
+        }
+
+        for (uint256 index = 0; index < users.length; index++) {
+            if (localVar.amounts[index] != 0) {
+                _supplyCApeForUser(
```
GopherJ, line 577: can we have common function defined for claimApeAndCompound and this function? seems too many duplicated code

market-config/auctionStrategies.ts
```diff
@@ -104,6 +104,16 @@ export const auctionStrategyUniswapV3: IAuctionStrategyParams = {
   tickLength: "900",
 };
 
+export const auctionStrategyBAKC: IAuctionStrategyParams = {
+  name: "auctionStrategyBAKC",
+  maxPriceMultiplier: utils.parseUnits("2.5", 18).toString(),
+  minExpPriceMultiplier: utils.parseUnits("1.2", 18).toString(),
+  minPriceMultiplier: utils.parseUnits("0.8", 18).toString(),
```
GopherJ, line 111: missing BAKC auction strategy https://docs.google.com/spreadsheets/d/1Yw2Cvj-QNfJ3N0N1ikFnxDvaieZeNGS-P_F1tSTUc7k/edit#gid=0

GopherJ:
seems the NTokenBAKC liquidation will not supply paired APE if the BAYC is not on our platform (paired on HorizonLabs), this may cause the collateral to escape from the system

GopherJ:
Shall SupplyLogic enable sAPE collateral flag for NTokenBAKC?

contracts/protocol/pool/PoolApeStaking.sol
GopherJ: `_supplyCApeForUser` may fail due to the precision loss of:

```
APE_COMPOUND.deposit(address(this), localVar.totalAmount);
```

WalidOfNow: @GopherJ  how would precision loss occur here?

contracts/protocol/pool/PoolApeStaking.sol
zhoujia6139: stack too deep

contracts/protocol/pool/PoolApeStaking.sol
zhoujia6139: @GopherJ  Do you have any data case. Precision loss is true, but I don't think it will cause _supplyCApeForUser fail.

contracts/protocol/pool/PoolApeStaking.sol
```diff
@@ -188,30 +180,33 @@ contract PoolApeStaking is
...
+                _nftPairs[index].bakcTokenId,
+                bakcNToken,
                 xTokenAddress,
-                _nftPairs[index].bakcTokenId
+                msg.sender
             );
         }
 
-        nTokenApeStaking.claimBAKC(_nftPairs, msg.sender);
+        INTokenApeStaking(xTokenAddress).claimBAKC(_nftPairs, msg.sender);
```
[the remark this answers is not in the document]
zhoujia6139: The cost is very little. and if we have a stack variable to store this value, it might cause stack too deep, then it'll cost more gas to use memory struct.

GopherJ (dismissed):

GopherJ (approved):

PR #316, opened 2023-01-10
P2P Pair Staking

zhoujia6139:
## Security Checklist

- [ ] 1. Re-Entrancy
- [ ] 3. Unexpected Ether
- [ ] 4. Delegatecall
- [ ] 5. Default Visibilities
- [ ] 6. Entropy Illusion

### ⚠️ NOTES ⚠️


contracts/misc/P2PPairStaking.sol
```diff
@@ -0,0 +1,586 @@
...
+        address _apeCoin,
+        address _cApe,
+        address _apeCoinStaking
+    ) {
+        bayc = _bayc;
+        mayc = _mayc;
+        bakc = _bakc;
+        apeCoin = _apeCoin;
+        cApe = _cApe;
+        apeCoinStaking = ApeCoinStaking(_apeCoinStaking);
```
GopherJ, line 60: shall we get bayc, mayc, bakc, apeCoin from apeCoinStaking? which may makes the deployment script easier

helpers/contracts-deployments.ts
```diff
@@ -2046,6 +2049,51 @@ export const deployAutoCompoundApe = async (verify?: boolean) => {
...
+    allTokens.BAYC.address,
+    allTokens.MAYC.address,
+    allTokens.BAKC.address,
+    allTokens.APE.address,
+    allTokens.cAPE.address,
+    apeCoinStaking,
+  ];
+
+  const p2pImplementation = await withSaveAndVerify(
+    new P2PPairStaking__factory(await getFirstSigner()),
```
GopherJ, line 2067: better split impl & proxy so that we can deploy impl for upgrading

scripts/deployments/steps/19_misc.ts
```diff
@@ -1,6 +1,7 @@
 import {
```
GopherJ, line 1: misc naming is too general I guess

contracts/misc/P2PPairStaking.sol
```diff
@@ -0,0 +1,586 @@
...
+                0
+            );
+        if (apeOrder.stakingType == StakingType.BAYCPairStaking) {
+            apeCoinStaking.depositBAKC(_stakingPairs, _otherPairs);
+        } else {
+            apeCoinStaking.depositBAKC(_otherPairs, _stakingPairs);
+        }
+
+        //5 emit event
+        emit PairStakingMatched(orderHash);
```
GopherJ, line 251: I think we need to distinguish two types of matching

contracts/misc/P2PPairStaking.sol
```diff
@@ -0,0 +1,586 @@
...
+
+        return orderHash;
+    }
+
+    function matchBAKCPairStakingList(
+        ListingOrder calldata apeOrder,
+        ListingOrder calldata bakcOrder,
+        ListingOrder calldata apeCoinOrder
+    ) external returns (bytes32 orderHash) {
```
GopherJ, line 175: shall we add matchingOperator check if msg.sender is not one of the three

contracts/misc/P2PPairStaking.sol
```diff
@@ -0,0 +1,586 @@
...
+        require(!isListingOrderCanceled[orderHash], "order already cancel");
+        isListingOrderCanceled[orderHash] = true;
+
+        emit OrderCancelled(orderHash, listingOrder.offerer);
+    }
+
+    function matchPairStakingList(
+        ListingOrder calldata apeOrder,
+        ListingOrder calldata apeCoinOrder
+    ) external returns (bytes32 orderHash) {
```
GopherJ, line 107: shall we add matchingOperator check if msg.sender is not one of the two

contracts/misc/P2PPairStaking.sol
```diff
@@ -0,0 +1,586 @@
...
+        //1 validate all order
+        _validateApeOrder(apeOrder);
+        _validateBakcOrder(bakcOrder);
+        _validateApeCoinOrder(apeCoinOrder);
+
+        //2 check if orders can match
+        require(
+            apeOrder.stakingType == StakingType.BAYCPairStaking ||
```
GopherJ, line 183: shall we be consistent on using `apeOrder.stakingType >= StakingType.BAYCPairStaking`

GopherJ:
we dont have non entrancy guard yet, better to add for important function calls

contracts/interfaces/IP2PPairStaking.sol
```diff
@@ -0,0 +1,156 @@
...
+        MAYCPairStaking
+    }
+
+    struct ListingOrder {
+        StakingType stakingType;
+        address offerer;
+        address token;
+        uint32 tokenId;
+        uint32 share;
+        uint256 startTime;
```
WalidOfNow, line 20: we can use smaller uint here. uint64 should be enough for timestamp

contracts/interfaces/IP2PPairStaking.sol
```diff
@@ -0,0 +1,156 @@
...
+        uint256 endTime;
```
WalidOfNow, line 21: same here

contracts/interfaces/IP2PPairStaking.sol
```diff
@@ -0,0 +1,156 @@
...
+        address apeToken;
+        address apeOfferer;
+        uint32 apeTokenId;
+        uint32 apeShare;
+        address bakcOfferer;
+        uint32 bakcTokenId;
+        uint32 bakcShare;
+        address apeCoinOfferer;
+        uint32 apeCoinShare;
+        uint256 apePrincipleAmount;
```
WalidOfNow, line 38: we can use uint128 here

contracts/interfaces/IP2PPairStaking.sol
```diff
@@ -0,0 +1,156 @@
...
+        uint8 v;
+        bytes32 r;
+        bytes32 s;
+    }
+
+    struct MatchedOrder {
+        StakingType stakingType;
+        address apeToken;
+        address apeOfferer;
+        uint32 apeTokenId;
```
WalidOfNow, line 31: should we use uint16 here? 2^16 is 65,536 which is larger than all token ids in BAYC, MAYC and BAKC

contracts/misc/P2PPairStaking.sol
```diff
@@ -0,0 +1,588 @@
...
+        );
+        uint256 apeAmount = getApeCoinStakingCap(apeOrder.stakingType);
+        IERC20(apeCoin).safeTransferFrom(
+            apeCoinOrder.offerer,
+            address(this),
+            apeAmount
+        );
+
+        //4 create match order
+        MatchedOrder memory matchedOrder;
```
WalidOfNow, line 144: is this more gas effecient than using `var = StructStype({param1: value, param2: value});` ?

contracts/interfaces/IP2PPairStaking.sol
zhoujia6139: ListingOrder is not stored on chain. so it's ok

contracts/interfaces/IP2PPairStaking.sol
zhoujia6139: apeOfferer, apeTokenId, apeShare will share a slot no matter apeTokenId is uint16 or uint32.

contracts/interfaces/IP2PPairStaking.sol
zhoujia6139: apePrincipleAmount will consume a single slot no matter it is uint128 or uint256

contracts/misc/P2PPairStaking.sol
zhoujia6139: I think it's ok. and i'll cost some gas to fetch this value. more over it has strong requirement for us to use the same bayc, mayc, bakc and apeCoin with apeCoinStaking.

contracts/misc/P2PPairStaking.sol
zhoujia6139: I think it's almost same for memory struct. For storage struct, it'll consume more gas in my way. Anyway, it's better practice, i'll change it.

contracts/misc/P2PPairStaking.sol
zhoujia6139: ok, it's more simpler. I'll change it.

contracts/misc/P2PPairStaking.sol
```diff
@@ -0,0 +1,618 @@
...
+        );
+        uint256 oldValue = compoundFee;
+        if (oldValue != _compoundFee) {
+            compoundFee = _compoundFee;
+            emit CompoundFeeUpdated(oldValue, _compoundFee);
+        }
+    }
+
+    function claimCompoundFee(address receiver) external onlyOwner {
+        this.claimCApeReward(receiver);
```
WalidOfNow, line 607: when we claim the compound fee, only owner can call it. but the function `claimCApeReward` claims the reward for `msg.sender` which will be the `owner()` and not `address(this)`we should probably make an internal function called _ claimCApeReward and send the the address of the claimer.

contracts/misc/P2PPairStaking.sol
```diff
@@ -0,0 +1,618 @@
...
+            apeCoinStaking.depositMAYC(singleNft);
+        }
+
+        //5 emit event
+        emit PairStakingMatched(orderHash);
+
+        return orderHash;
+    }
+
+    function matchBAKCPairStakingList(
```
WalidOfNow, line 185: There's a lot of redundancy in the code between `matchPairStakingList` and `matchBAKCPairStakingList`. is it possible to abstract that redundancy?

contracts/misc/P2PPairStaking.sol
```diff
@@ -47,6 +47,7 @@ contract P2PPairStaking is
     bytes32 public DOMAIN_SEPARATOR;
     mapping(bytes32 => bool) public isListingOrderCanceled;
     mapping(bytes32 => MatchedOrder) public matchedOrders;
+    mapping(uint256 => uint256) public apeMatchedCount;
```
yubo-ruan, line 50: 这个地方是不是太贵了，我们用mapping(uint256 => uint8)就够了？

contracts/misc/P2PPairStaking.sol
```diff
@@ -0,0 +1,688 @@
...
+            }
+        }
+        //5 transfer token
+        uint256 matchedCount = apeMatchedCount[order.apeToken][
+            order.apeTokenId
+        ];
+        if (matchedCount == 1) {
+            IERC721(order.apeToken).safeTransferFrom(
+                address(this),
+                _getApeNTokenAddress(order.apeToken),
```
i5possible, line 338:
```suggestion
                apeNToken,
```

contracts/misc/P2PPairStaking.sol
```diff
@@ -0,0 +1,688 @@
...
+    address public immutable bakc;
+    address public immutable nBayc;
+    address public immutable nMayc;
+    address public immutable nBakc;
+    address public immutable apeCoin;
+    address public immutable cApe;
+    ApeCoinStaking public immutable apeCoinStaking;
+
+    bytes32 public DOMAIN_SEPARATOR;
+    mapping(bytes32 => bool) public isListingOrderCanceled;
```
GopherJ, line 51: typo here, Cancelled

contracts/misc/P2PPairStaking.sol
```diff
@@ -0,0 +1,688 @@
...
+    bytes32 public constant MATCHED_ORDER_HASH =
+        keccak256(
+            "MatchedOrder(uint8 stakingType,address apeToken,uint32 apeTokenId,uint32 apeShare,uint32 bakcTokenId,uint32 bakcShare,address apeCoinOfferer,uint32 apeCoinShare,uint256 apePrincipleAmount)"
+        );
+    bytes32 internal constant EIP712_DOMAIN =
+        keccak256(
+            "EIP712Domain(string name,string version,uint256 chainId,address verifyingContract)"
+        );
+
+    address public immutable bayc;
```
GopherJ, line 40: shall we make them private to reduce contract size?

GopherJ (approved):

PR #351, opened 2023-03-18
Fix hack record

zhoujia6139:
## Security Checklist

- [ ] 1. Re-Entrancy
- [ ] 3. Unexpected Ether
- [ ] 4. Delegatecall
- [ ] 5. Default Visibilities
- [ ] 6. Entropy Illusion

### ⚠️ NOTES ⚠️

### DIFF
--- a/contracts/misc/AutoCompoundApe.sol
+++ b/contracts/misc/AutoCompoundApe.sol
@@ -138,8 +138,11 @@ contract AutoCompoundApe is
             0
         );
         if (rewardAmount > 0) {
+            uint256 balanceBefore = apeCoin.balanceOf(address(this));
             apeStaking.claimSelfApeCoin();
-            bufferBalance += rewardAmount;
+            uint256 balanceAfter = apeCoin.balanceOf(address(this));
+            uint256 realClaim = balanceAfter - balanceBefore;
+            bufferBalance += realClaim;
         }
     }
 
@@ -187,6 +190,10 @@ contract AutoCompoundApe is
 
     function unpause() external onlyOwner {
         _unpause();
+    }
+
+    function rebaseFromApeCoinStaking() external onlyOwner {
+        (stakingBalance, ) = apeStaking.addressPosition(address(this));
     }
 
     /**
--- a/contracts/misc/AutoCompoundApe.sol
+++ b/contracts/misc/AutoCompoundApe.sol
@@ -179,6 +179,14 @@ contract AutoCompoundApe is
         return IDelegation(delegateContract).delegation(address(this), spaceId);
     }
 
+    function pasue() external onlyOwner {
+        _pause();
+    }
+
+    function unpasue() external onlyOwner {
+        _unpause();
+    }    
+
     function withdrawFromApeCoinStaking(address receiver) external onlyOwner {
         uint256 amount = 2332214464588784613678467;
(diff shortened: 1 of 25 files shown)


contracts/misc/AutoCompoundApe.sol
```diff
@@ -29,6 +30,7 @@ contract AutoCompoundApe is
     ApeCoinStaking public immutable apeStaking;
     IERC20 public immutable apeCoin;
     uint256 public bufferBalance;
+    uint256 public bufferStakingBalance;
```
i5possible, line 33:
```suggestion
    uint256 public stakingBalance;
```
It's not a buffer that will be cleared. So maybe we use stakingaBalance or stakingAmount directly?

contracts/protocol/libraries/logic/SupplyLogic.sol
```diff
@@ -69,6 +70,81 @@ library SupplyLogic {
...
+    }
+
+    function executeWithdrawUserPosition(
+        mapping(address => DataTypes.ReserveData) storage reservesData,
+        address user,
+        address asset
+    ) external {
+        DataTypes.ReserveData storage currentReserve = reservesData[asset];
+        address xTokenAddress = currentReserve.xTokenAddress;
+        uint256 liquidityIndex = currentReserve.liquidityIndex;
```
GopherJ, line 141: we didn't updateState here

contracts/protocol/libraries/logic/SupplyLogic.sol
```diff
@@ -69,6 +70,81 @@ library SupplyLogic {
...
+        uint256 pBalance = IPToken(xTokenAddress).balanceOf(user);
+        if (pBalance > 0) {
+            IPToken(xTokenAddress).burn(user, user, pBalance, liquidityIndex);
```
GopherJ, line 144: we may want to just burn ptoken without sending assets to user, so shall we define receiver in the parameters

contracts/misc/CApe.sol
```diff
@@ -348,6 +349,7 @@ abstract contract CApe is ContextUpgradeable, ICApe, PausableUpgradeable {
         address _recipient,
         uint256 _sharesAmount
     ) internal whenNotPaused {
+        revert("paused");
```
GopherJ, line 352: we still need to deposit APE as cAPE then supply to lending right?

contracts/protocol/libraries/logic/ValidationLogic.sol
```diff
@@ -388,20 +388,20 @@ library ValidationLogic {
             DataTypes.AssetType assetType
         ) = reserveCache.reserveConfiguration.getFlags();
         require(isActive, Errors.RESERVE_INACTIVE);
-        require(!isPaused, Errors.RESERVE_PAUSED);
+        // require(!isPaused, Errors.RESERVE_PAUSED);
```
GopherJ, line 391: user will be able to repay, is it expected?

contracts/misc/AutoCompoundApe.sol
zhoujia6139: ok

contracts/protocol/libraries/logic/SupplyLogic.sol
zhoujia6139: yeah.

PR #353, opened 2023-03-20
Timelock Implementation

WalidOfNow:
## Security Checklist

- [ ] 1. Re-Entrancy
- [ ] 3. Unexpected Ether
- [ ] 4. Delegatecall
- [ ] 5. Default Visibilities
- [ ] 6. Entropy Illusion

### ⚠️ NOTES ⚠️


contracts/misc/DefaultTimeLockStrategy.sol
```diff
@@ -0,0 +1,44 @@
+// SPDX-License-Identifier: MIT
+pragma solidity 0.8.10;
+
+import "../interfaces/ITimeLockStrategy.sol";
+
+contract MockTestTimeLockStrategy is ITimeLockStrategy {
```
jfzhou5, line 6: MockTestTimeLockStrategy => DefaultTimeLockStrategy

contracts/misc/TimeLock.sol
```diff
@@ -0,0 +1,142 @@
...
+    event AgreementClaimed(
+        uint256 agreementId,
+        DataTypes.AssetType assetType,
+        address indexed asset,
+        uint256[] tokenIdsOrAmounts,
+        address indexed beneficiary
+    );
+    mapping(uint256 => Agreement) public agreements;
+
+    uint248 public agreementCount;
```
jfzhou5, line 45: Should we update uint248 to uint256?

contracts/protocol/libraries/logic/LiquidationLogic.sol
```diff
@@ -451,13 +451,15 @@ library LiquidationLogic {
...
+        // no time lock needed here
+        DataTypes.TimeLockParams memory timeLockParams;
         // Burn the equivalent amount of xToken, sending the underlying to the liquidator
         IPToken(vars.collateralXToken).burn(
             params.borrower,
             params.liquidator,
             vars.actualCollateralToLiquidate,
-            collateralReserveCache.nextLiquidityIndex
+            collateralReserveCache.nextLiquidityIndex,
+            timeLockParams
```
GopherJ, line 462: we are injecting the logic into each burn + transferUnderlying except liquidation. it's a lot of changes and hard to be consistent, can we make them all into PToken?

contracts/misc/TimeLock.sol
WalidOfNow: it's `uint248` to save storage and get packed with the state variable after it into one slot

jfzhou5: Cool, understood.

contracts/misc/TimeLock.sol
```diff
@@ -0,0 +1,145 @@
...
+    );
+
+    event AgreementClaimed(
+        uint256 agreementId,
+        DataTypes.AssetType assetType,
+        address indexed asset,
+        uint256[] tokenIdsOrAmounts,
+        address indexed beneficiary
+    );
+    mapping(uint256 => Agreement) public agreements;
```
jfzhou5, line 44: @WalidOfNow The public view function `agreements(id)` just returns (assetType, actionType, isFrozen, asset, beneficiary, releaseTime), not contains `tokenIdsOrAmounts `, you need add another view function for us to get the  
 full Agreement data.

yubo-ruan: View function can be added tomorrow.  Let's have the PR ready for secure3 first I think

jfzhou5: Yeah, I agree.

contracts/misc/DefaultTimeLockStrategy.sol
```diff
@@ -0,0 +1,85 @@
...
+    function resetPeriodLimit() internal {
+        totalAmountInCurrentPeriod = 0;
+        lastResetTimestamp = uint48(block.timestamp);
+    }
+
+    function calculateTimeLockParams(
+        DataTypes.TimeLockFactorParams calldata params
+    ) external onlyPool returns (DataTypes.TimeLockParams memory) {
+        uint256 currentTimestamp = block.timestamp;
+        if (currentTimestamp - lastResetTimestamp >= PERIOD) {
```
nzhl, line 61: what if user undertakes withdraw right before the reset point and undertake again right after. Should this be allowed ?

WalidOfNow: yeah they can do that for now. this is just V1 so we understand that such behaviour is possible and accept the risk

contracts/misc/TimeLock.sol
```diff
@@ -0,0 +1,197 @@
...
+                        address(this),
+                        agreement.beneficiary,
+                        agreement.tokenIdsOrAmounts[i]
+                    );
+                }
+            }
+        }
+    }
+
+    function claimMoonBirds(uint256 agreementId) external nonReentrant {
```
nzhl, line 161: are we having moonbird claim in batch as well ?

contracts/misc/TimeLock.sol
```diff
@@ -0,0 +1,187 @@
...
+            Agreement memory agreement = _validateAndDeleteAgreement(
+                agreementIds[index]
+            );
+
+            if (agreement.assetType == DataTypes.AssetType.ERC20) {
+                IERC20(agreement.asset).safeTransfer(
+                    agreement.beneficiary,
+                    agreement.tokenIdsOrAmounts[0]
+                );
+            } else if (agreement.assetType == DataTypes.AssetType.ERC721) {
```
infinityhacker, line 134: not allow to claim moobird nft in this else logic, or the corresponding agreemenIds will be deleted

contracts/protocol/libraries/logic/BorrowLogic.sol
```diff
@@ -99,9 +101,21 @@ library BorrowLogic {
         );
 
         if (params.releaseUnderlying) {
+            DataTypes.TimeLockParams memory timeLockParams = GenericLogic
+                .calculateTimeLockParams(
+                    reserve,
+                    DataTypes.TimeLockFactorParams({
+                        assetType: DataTypes.AssetType.ERC20,
+                        asset: params.asset,
+                        amount: params.amount
```
nzhl, line 110: We should not using params.amout here because -1 (i.e. uint256.max)  is possiblely passed in for denoting withdraw all. In this case locked time returned by `calculateTimeLockParams ` might be wrong.

nzhl: Also it will cause `calculateTimeLockParams` revert because there is a explicit conversion from u256 to u128
```
        timeLockParams.releaseTime +=
            currentTimestamp +
            _updatePeriodLimit(currentTimestamp, uint128(params.amount));
```

PR #370, opened 2023-04-14
feat: raffle impl

WalidOfNow:
## Security Checklist

- [ ] 1. Re-Entrancy
- [ ] 3. Unexpected Ether
- [ ] 4. Delegatecall
- [ ] 5. Default Visibilities
- [ ] 6. Entropy Illusion

### ⚠️ NOTES ⚠️

### DIFF
--- a/contracts/misc/ParaSpaceRaffle.sol
+++ b/contracts/misc/ParaSpaceRaffle.sol
@@ -6,8 +6,6 @@ import {MerkleProof} from "../dependencies/openzeppelin/contracts/MerkleProof.so
 import {Ownable} from "../dependencies/openzeppelin/contracts/Ownable.sol";
 
 contract ParaSpaceRaffle is VRFV2WrapperConsumerBase, Ownable {
-    address immutable linkAddress;
-    address immutable wrapperAddress;
     bytes32 public immutable merkleRoot;
     uint256 immutable numCandidates;
 
@@ -24,8 +22,6 @@ contract ParaSpaceRaffle is VRFV2WrapperConsumerBase, Ownable {
         uint256 _numCandidates,
         bytes32 _merkleRoot
     ) VRFV2WrapperConsumerBase(_linkAddress, _wrapperAddress) {
-        linkAddress = _linkAddress;
-        wrapperAddress = _wrapperAddress;
         numCandidates = _numCandidates;
         merkleRoot = _merkleRoot;
     }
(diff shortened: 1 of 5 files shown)


contracts/misc/ParaSpaceRaffle.sol
```diff
@@ -0,0 +1,83 @@
+pragma solidity ^0.8.10;
+
+import {VRFV2WrapperConsumerBase} from "../dependencies/chainlink/VRFV2WrapperConsumerBase.sol";
+import {LinkTokenInterface} from "../dependencies/chainlink/interfaces/LinkTokenInterface.sol";
+import {MerkleProof} from "../dependencies/openzeppelin/contracts/MerkleProof.sol";
+import {Ownable} from "../dependencies/openzeppelin/contracts/Ownable.sol";
+
+contract ParaSpaceRaffle is VRFV2WrapperConsumerBase, Ownable {
```
GopherJ, line 8: can we use ACLManager instead to avoid managing so many owners

helpers/contracts-deployments.ts
```diff
@@ -3121,3 +3123,12 @@ export const deployDelegationRegistry = async (verify?: boolean) =>
     [],
     verify
   ) as Promise<DelegationRegistry>;
+
+
+  export const deployParaSapaceRaffle = async (_linkAddress, _wrapperAddress, numCandidates, tree, verify?: boolean) =>
+  withSaveAndVerify(
+    new ParaSpaceRaffle__factory(await getFirstSigner()),
+    eContractid.DelegationRegistry,
```
GopherJ, line 3131: typo here?

contracts/misc/ParaSpaceRaffle.sol
```diff
@@ -0,0 +1,83 @@
...
+    address immutable linkAddress;
+    address immutable wrapperAddress;
```
zhoujia6139, line 10: Seems there is no need for this variable.

contracts/misc/ParaSpaceRaffle.sol
```diff
@@ -0,0 +1,83 @@
...
+
+    function _requestRandomness(uint16 _wordsNum) internal {
+        requestRandomness(
+            uint32(callbackGasLimit * _wordsNum),
+            requestConfirmations,
+            _wordsNum
+        );
+    }
+
+    function verifyWinner(uint256 index, bytes32[] calldata merkleProof)
```
GopherJ, line 67: why dont we check if winners[index] is true and why this function is useful onchain?

contracts/misc/ParaSpaceRaffle.sol
WalidOfNow: this is a one time contract. no need for that

PR #379, opened 2023-04-20
Otherdeed Flashclaim

WalidOfNow:
## Security Checklist

- [ ] 1. Re-Entrancy
- [ ] 3. Unexpected Ether
- [ ] 4. Delegatecall
- [ ] 5. Default Visibilities
- [ ] 6. Entropy Illusion

### ⚠️ NOTES ⚠️


contracts/protocol/libraries/logic/PositionMoverLogic.sol
```diff
@@ -148,4 +157,127 @@ library PositionMoverLogic {
...
+            if (newCollateralizedBalance == 0) {
+                userConfig.setUsingAsCollateral(reserve.id, false);
+                emit ReserveUsedAsCollateralDisabled(
+                    params.otherdeed,
+                    msg.sender
+                );
+            }
+        }
+
+        IERC721(params.otherdeed).setApprovalForAll(params.vessel, true);
```
WalidOfNow, line 202: should we do this only once? I made it this way for security reason

contracts/protocol/libraries/logic/PositionMoverLogic.sol
```diff
@@ -148,4 +157,127 @@ library PositionMoverLogic {
...
+        );
+
+        DataTypes.TimeLockParams memory timeLockParams;
+
+        (
+            uint64 oldCollateralizedBalance,
+            uint64 newCollateralizedBalance
+        ) = INToken(reserveCache.xTokenAddress).burn(
+                msg.sender,
+                address(this),
```
GopherJ, line 184: shouldn't this be reserveCache.xTokenAddress

contracts/protocol/libraries/logic/PositionMoverLogic.sol
```diff
@@ -148,4 +157,127 @@ library PositionMoverLogic {
...
+        }
+
+        IERC721(params.otherdeed).setApprovalForAll(params.vessel, false);
+
+        SupplyLogic.executeSupplyERC721(
+            reservesData,
+            userConfig,
+            DataTypes.ExecuteSupplyERC721Params({
+                asset: params.otherdeedExpanded,
+                tokenData: _getTokenData(params.otherdeedIds),
```
GopherJ, line 222: if there is any otherdeedExpanded... stuck in our pool. The others can get them

WalidOfNow: how would that happen? the `claimVessels` would fail if the caller doesn't own the the otherdeeds.

contracts/protocol/libraries/logic/PositionMoverLogic.sol
```diff
@@ -148,4 +157,127 @@ library PositionMoverLogic {
...
+            uint64 oldCollateralizedBalance,
+            uint64 newCollateralizedBalance
+        ) = INToken(reserveCache.xTokenAddress).burn(
+                msg.sender,
+                address(this),
+                params.otherdeedIds,
+                timeLockParams
+            );
+
+        bool isWithdrawCollateral = (newCollateralizedBalance <
```
zhoujia6139, line 189: I think we don't need this `isWithdrawCollateral`, just need to check if `newCollateralizedBalance == 0`, then we can call `setUsingAsCollateral`

contracts/protocol/libraries/logic/PositionMoverLogic.sol
GopherJ: never mind, I was thinking using wrong vessel address, but since we need to supply. I think it's ok

PR #400, opened 2023-07-27
Account abstraction

WalidOfNow:
## Security Checklist

- [ ] 1. Re-Entrancy
- [ ] 3. Unexpected Ether
- [ ] 4. Delegatecall
- [ ] 5. Default Visibilities
- [ ] 6. Entropy Illusion

### ⚠️ NOTES ⚠️


contracts/account-abstraction/Account.sol
```diff
@@ -0,0 +1,402 @@
...
+        uint256 value,
+        bytes calldata data
+    ) external payable onlyAuthorized returns (bytes memory) {
+        emit TransactionExecuted(to, value, data);
+
+        _incrementNonce();
+
+        return _call(to, value, data);
+    }
+
```
jfzhou5, line 109: Can we add a batchExecuteCall function to make the batch call easy? 
Like this https://github.com/eth-infinitism/account-abstraction/blob/develop/contracts/samples/SimpleAccount.sol#L66

contracts/account-abstraction/Account.sol
```diff
@@ -128,9 +127,9 @@ contract Account is
     ) internal virtual override returns (uint256 validationData) {
         bytes32 hash = userOpHash;
 
-        if (address.code.length == 0) {
-            hash = userOpHash.toEthSignedMessageHash();
-        }
+//        if (address.code.length == 0) {
```
billyz9611, line 130: do you want this: address(owner).code.length

PR #402, opened 2023-08-14
Izumi lp integration

zhoujia6139:
## Security Checklist

- [ ] 1. Re-Entrancy
- [ ] 3. Unexpected Ether
- [ ] 4. Delegatecall
- [ ] 5. Default Visibilities
- [ ] 6. Entropy Illusion

### ⚠️ NOTES ⚠️


contracts/protocol/tokenization/NTokenLiquidity.sol
```diff
@@ -0,0 +1,232 @@
...
+        (uint256 amount0, uint256 amount1) = _increaseLiquidity(
+            positionManager,
+            tokenId,
+            amountAdd0,
+            amountAdd1,
+            amount0Min,
+            amount1Min
+        );
+
+        // refund unused tokens
```
billyz9611, line 180: if token0 & token1 are not WETH, we need to refund ETH to user when msg.value > 0.

contracts/protocol/tokenization/NTokenLiquidity.sol
```diff
@@ -0,0 +1,204 @@
...
+
+        // move underlying into this contract
+        bool token0IsETH;
+        bool token1IsETH;
+        if (msg.value > 0) {
+            address weth = _addressesProvider.getWETH();
+            require(token0 == weth || token1 == weth, Errors.INVALID_AMOUNT);
+            token0IsETH = (token0 == weth);
+            token1IsETH = (token1 == weth);
+        }
```
GopherJ, line 143: there are many duplicated code here, I think we can declare variable to represent ethToken, ethAmount

contracts/protocol/tokenization/NTokenLiquidity.sol
```diff
@@ -0,0 +1,204 @@
...
+        // refund unused tokens
+        if (amount0 < amountAdd0) {
+            uint256 refund0 = amountAdd0 - amount0;
+            if (token0IsETH) {
+                _refundETH(positionManager);
+                Helpers.safeTransferETH(payer, refund0);
+            } else {
+                IERC20(token0).safeTransfer(payer, refund0);
+            }
+        }
```
GopherJ, line 175: same here I think we can remove duplicated code

GopherJ:
I think upgrade script should also be fixed by adding Izumi LP

contracts/interfaces/IXTokenType.sol
```diff
@@ -25,6 +25,7 @@ enum XTokenType {
     NTokenOtherdeed,
     NTokenStakefish,
     NTokenChromieSquiggle,
+    NTokenIZUMILp,
```
GopherJ, line 28: this will break PTokenStKSM which is deployed on moonbeam

PR #413, opened 2023-09-03
Borrow and stake support pcape

zhoujia6139:
## Security Checklist

- [ ] 1. Re-Entrancy
- [ ] 3. Unexpected Ether
- [ ] 4. Delegatecall
- [ ] 5. Default Visibilities
- [ ] 6. Entropy Illusion

### ⚠️ NOTES ⚠️


contracts/protocol/pool/PoolApeStaking.sol
```diff
@@ -305,11 +314,38 @@ contract PoolApeStaking is
...
+                DataTypes.ReserveCache memory cApeReserveCache = cApeReserve
+                    .cache();
+                cApeReserve.updateState(cApeReserveCache);
+                cApeReserve.updateInterestRates(
+                    cApeReserveCache,
+                    address(APE_COMPOUND),
+                    0,
+                    stakingInfo.cashAmount
+                );
+                IPToken(PCAPE).burn(
```
GopherJ, line 336: shall we check if we need to turn off collateral?

GopherJ, line 336: since we already read cApeReserveCache, why not we just use xTokenAddress from that?

helpers/contracts-deployments.ts
```diff
@@ -553,12 +553,21 @@ export const deployPoolApeStaking = async (
 
   const config = getParaSpaceConfig();
   const treasuryAddress = config.Treasury;
+
+  const cApe = await getAutoCompoundApe();
+  let pcApeAddress = "0x5F56386F21766237707a447FB37C717dC221bb13"; //for test env
```
GopherJ, line 558: this breaks other deployment

contracts/protocol/pool/PoolApeStaking.sol
zhoujia6139: yeah. We should.

zhoujia6139: yes，we can.

contracts/protocol/pool/PoolApeStaking.sol
```diff
@@ -304,12 +164,57 @@ contract PoolApeStaking is
...
+                );
+            } else {
+                //pcApe
+                DataTypes.ReserveData storage cApeReserve = ps._reserves[
+                    address(APE_COMPOUND)
+                ];
+                DataTypes.ReserveCache memory cApeReserveCache = cApeReserve
+                    .cache();
+                address PCAPE = cApeReserveCache.xTokenAddress;
+                require(
```
billyz9611, line 185: suggest require user's pcApe balance >=  cashAmount

contracts/protocol/pool/PoolApeStaking.sol
```diff
@@ -304,12 +164,57 @@ contract PoolApeStaking is
...
+                    stakingInfo.cashAsset == PCAPE,
+                    Errors.INVALID_ASSET_TYPE
+                );
+                cApeReserve.updateState(cApeReserveCache);
```
billyz9611, line 189: suggest check cAPE status is not Paused * is Active

